PR #3, opened 2019-06-21
Travis support + uses global elm programs

harrysarson:
Hi!

This pull request adds CI support, once it is enabled at https://travis-ci.com/account/repositories.

To get the CI passing I had to use globally installed elm and elm-test programs (see strange error in https://travis-ci.com/harrysarson/compiler-1/builds/116453794). If the `node_modules/.bin/*` approach is one that you are keen to keep then I am more than happy to have another crack at the strange error. :)

Refs: https://travis-ci.com/harrysarson/compiler-1


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+
+node_js:
+  lts/*
+
+before_install:
+  - npm install -g elm@0.19.0 elm-test@0.19.0 elm-format@0.8 elm-analyse@0.16
```
Janiczek: Can we make the versions here same as the versions that were in `package.json`? (Or most recent)

Also, `elm-format` and `elm-analyse` aren't used here nor in the Makefile - can we remove them from here and from README? (I only had them here because my personal habit is to use tools from `node_modules/.bin/`)

README.md
```diff
@@ -85,7 +87,7 @@ Runs `elm-test` on the test suite (gasp!)
 - [ ] Create issues for all the TODOs in the codebase, tag some with <kbd>help wanted</kbd> and <kbd>good first issue</kbd>
 - [ ] [@janiczek](https://twitter.com/janiczek/): Write some more TODOs/issues off the top of your head, and maybe some of the long-term plans
 - [ ] [@janiczek](https://twitter.com/janiczek/): Share your Firefox bookmarks relevant to `elm-in-elm` (ie. talks about Haskell hierarchical optimizations etc.)
-- [ ] Travis integration + shields.io badge about it
+- [x] Travis integration + shields.io badge about it
```
Janiczek: Feel free to add yourself to the Contributors table here in the README :slightly_smiling_face:

Janiczek:
Hi! :smile: Wow, didn't expect contributions that fast!

So, the PR is great - just a few things I've commented above.

I'd like `package.json` and `yarn`/`npm` solution better, but I guess we can always look at that later. This is great on its own :slightly_smiling_face:

README.md
```diff
@@ -60,9 +62,9 @@ The tooling around this project requires:
 
 * **`make`** for the [Makefile](Makefile)
 * **NodeJS 10+** for the [CLI tool](src/index.js)
-* **`yarn`** for installing `elm` and `elm-test` (among other dependencies from [package.json](package.json))
+* **`elm`**, **`elm-test`**, **`elm-format`** and **`elm-analyse`**
```
Janiczek: Let's only mention `elm` and `elm-test` here.

README.md
```diff
@@ -6,6 +6,8 @@
 # elm-in-elm
 Elm compiler written in Elm!
 
+<a href="https://travis-ci.com/elm-in-elm/compiler" alt="Build Status">
+  <img src="https://travis-ci.com/elm-in-elm/compiler.svg?branch=master" /></a>
```
Janiczek: Oh, also one more thing: let's use the shields.io badge:
```
https://img.shields.io/travis/com/elm-in-elm/compiler/master.svg
```

The original one has a bit different style (at least for me in Firefox@Linux) so if we can, let's make it consistent :)

.travis.yml
harrysarson: > Can we make the versions here same as the versions that were in package.json? (Or most recent)

Yes, I was attempting to install latest `0.19.0*` version but did not get it quite right.

> Also, elm-format and elm-analyse aren't used here nor in the Makefile - can we remove them from here and from README? (I only had them here because my personal habit is to use tools from node_modules/.bin/)

Yep :)

Janiczek:
Great stuff, thank you @harrysarson !

PR #4, opened 2019-06-22
replace why/not why with goals/non goals

harrysarson:
I had to read the text for the `## False "Why?"` section to understand what it meant
Possibly `## Non goals` is clearer and it is a pattern I have seen used in other projects.

A subjective change, please take or leave.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Elm compiler written in Elm!
 
 ----
 
-## Why?
+## Goals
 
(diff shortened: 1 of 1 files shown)


Janiczek:
Yeah, seems better!

The reason I started with "Why" and then added "False why" is because of some "awesome README examples" where there was "Always start with Why?" :smile: 

"Non goals" seem better to me than "False why", so let's do your proposed change :+1:

PR #9, opened 2019-06-28
add unit type

### LINKED ISSUE #8: Implement unit value and type
`()` should be a valid Elm syntax. What we emit it as is probably not that important right now (we will find out later!) - let's emit as `{"type": "unit"}`

That means we need to implement:
- [ ] parser tests
- [ ] emit tests (if done after #7 )
- [ ] new AST.Frontend.Expr case
- [ ] parser
- [ ] new type (treat it similarly to literals in type inference)
- [ ] JS emit case

No optimizations needed (or at least I can't think of any), no desugaring needed either.

The compiler errors will probably guide you :sweat_smile: Ask here or on Discord if anything is unclear!

Warry:
fixes #8

### DIFF
--- a/src/compiler/Error.elm
+++ b/src/compiler/Error.elm
@@ -53,6 +53,7 @@ type ParseContext
     | InLet
     | InLetBinding
     | InLambda
+    | InUnit
 
 
 type ParseProblem
@@ -77,6 +78,7 @@ type ParseProblem
     | ExpectingEqualsSign -- `x >=< 1`
     | ExpectingMinusSign -- `>-<42`
     | ExpectingInt
+    | ExpectingHexadecimals
     | ExpectingSingleQuote
     | ExpectingChar
     | ExpectingDoubleQuote
@@ -96,6 +98,7 @@ type ParseProblem
     | ExpectingIn
     | ExpectingAtLeastOne
     | ExpectingNewline
+    | ExpectingUnit
     | InvalidInt
     | CompilerBug String
 
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -63,6 +63,8 @@ emitExpr ( expr, type_ ) =
             in
             "(() => {" ++ bindingsJS ++ "; return " ++ emitExpr body ++ ";})()"
 
+        Unit ->
+            """{ "type": "unit" }"""
 
 {-| TODO test
 -}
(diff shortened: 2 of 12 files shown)


src/compiler/Error.elm
```diff
@@ -53,6 +53,7 @@ type ParseContext
     | InLet
     | InLetBinding
     | InLambda
+    | InUnit
```
Warry: this looks useless, wdyt?

Janiczek: I believe it can help us when debugging other parsers - I'd keep it. (But, I have to get up to speed with "how best to debug parsers" myself. That's the part that gives me the most trouble when writing them :) )

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -63,6 +63,8 @@ emitExpr ( expr, type_ ) =
             in
             "(() => {" ++ bindingsJS ++ "; return " ++ emitExpr body ++ ";})()"
 
+        Unit ->
+            "{ \"type\": \"unit\" }"
```
Janiczek: we can probably get rid of the spaces and quotes now that I look at it
```elm
"""{type: "unit"}"""
```

tests/ParserTest.elm
```diff
@@ -596,6 +596,14 @@ expr =
                   )
                 ]
               )
+            , ( "unit is unit"
+              , [ ( "unit"
```
Janiczek: Let's keep the outer description for the expression case (in this case `"unit"`) and the inner one for what is the specific example testing. In this case, there's not much to test, so I'd just write `"simple case"`.

Janiczek:
@Warry Only small nitpicks in the comments, great stuff!

And two more things:
:one: Could you please change the table in the roadmap? There's `unit type` there. As far as I'm concerned, all the columns of that row can be `:heavy_check_mark:` :heavy_check_mark: 
:two: And could you please also (if you want to!) add yourself to the Maintainers table in README? Your GitHub user ID (judging by avatar URL) is `197573` so the URL would be something like `https://avatars2.githubusercontent.com/u/197573?v=3&s=150`

Janiczek:
Actually, because we merged emit tests first, let's also add an emit test for Unit into [EmitTest.elm](tests/EmitTest.elm) please

Warry:
this time, it should be okay!

Janiczek:
@Warry build fails because it's missing `InUnit` - after adding that I can merge this :) thanks!

PR #13, opened 2019-06-29
Add tests for Int, including hexadecimals

Warry:


Janiczek:
It would be a good test (or a few) to have - fail parsing if hex int is longer than N... fail parsing if normal int is above some MAX_INT value... etc

Warry:
I hesitated, but there were no other failing test. I'll patch this (using Number.MAX_SAFE_INTEGER, instead of MAX_INTEGER, which depends on the platform).

Janiczek:
On a second thought, this is probably platform-dependent. Let's keep this how it is for now. I'm not sure how to proceed with arbitrary limits - whether to enforce them too or to extend past them. Let's postpone that decision :)

I'm merging this as is. Thank you @Warry!

PR #15, opened 2019-06-30
add literal char tests

### LINKED ISSUE #11: Write comprehensive parser tests for characters
This means:

- [ ] find all the ways characters can be written in Elm (`\u{...}`? `\x{...}`? straight-up emoji in the source code? something else?)
- [ ] add those as tests to the "literal char" section of ParserTest.elm so that the tests exercise all the possibilities and truthfully tell us what's done and what's not
- [ ] update the "what's done" table in README - parser tests for characters

aaronjanse:
fix #11

I don't know if this is "comprehensive," but it's a start.

### DIFF
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -548,7 +548,7 @@ expr =
                   , Ok (Literal (Char '😃'))
                   )
                 , ( "escaped single quote"
-                  , "'\''"
+                  , "'\\\''"
                   , Ok (Literal (Char '\''))
                   )
                 , ( "emoji code point"
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -543,6 +543,10 @@ expr =
                   , "'A'"
                   , Ok (Literal (Char 'A'))
                   )
+                , ( "double quote"
+                  , "'\"'"
+                  , Ok (Literal (Char '\"'))
+                  )
                 , ( "emoji"
                   , "'😃'"
                   , Ok (Literal (Char '😃'))
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -543,6 +543,18 @@ expr =
                   , "'A'"
                   , Ok (Literal (Char 'A'))
                   )
+                , ( "emoji"
+                  , "'😃'"
+                  , Ok (Literal (Char '😃'))
+                  )
+                , ( "escaped single quote"
+                  , "'\''"
+                  , Ok (Literal (Char '\''))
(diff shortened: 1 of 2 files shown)


aaronjanse:
I fixed the escaped quote test. It now fails, unfortunately:
```
✗ escaped single quote

    '\''
    ===>
    Err
    
    
        (0,4) ExpectingChar
        ---- with context stack ----
        (0,3) InLiteral
        (0,0) InExpr
```

I'm basing the chars off of the docs on [Char in elm/core](https://package.elm-lang.org/packages/elm/core/latest/Char#Char).

> ```elm
> 'a'
> '0'
> 'Z'
> '?'
> '"'
> 'Σ'
> '🙈'
> '\t'
> '\"'
> '\''
> '\u{1F648}' -- '🙈'
> ```

We might want to test every escaped character because there aren't many.

Janiczek:
> We might want to test every escaped character because there aren't many.

I agree, let's do that!

> I fixed the escaped quote test. It now fails, unfortunately:

Good!

tests/ParserTest.elm
```diff
@@ -543,6 +543,22 @@ expr =
                   , "'A'"
                   , Ok (Literal (Char 'A'))
                   )
+                , ( "double quote"
+                  , "'\"'"
+                  , Ok (Literal (Char '\"'))
```
Janiczek: the double-quote char on the `Ok ...` line doesn't need to be escaped, but that's just a nitpick :slightly_smiling_face:

tests/ParserTest.elm
```diff
@@ -543,6 +543,22 @@ expr =
...
+                  )
+                , ( "emoji"
+                  , "'😃'"
+                  , Ok (Literal (Char '😃'))
+                  )
+                , ( "escaped single quote"
+                  , "'\\\''"
```
Janiczek: I wondered if `"""'\''"""` wouldn't be more readable, but ... probably not :upside_down_face:

Janiczek:
You'll be basing this off the Parse.String in the Elm compiler?
https://github.com/elm/compiler/blob/master/compiler/src/Parse/String.hs#L279-L285

JavaScript seems to have more: https://en.wikipedia.org/wiki/Escape_character#JavaScript
but those (eg. `\b` or `\xFF`) aren't supported in Elm...

aaronjanse:
Thanks for the suggestions!

I'm close to a working char parser escapes implementation. Once I'm done, I'll submit that code along with the better testing code you suggested.

PR #31, opened 2019-07-04
Add literal string tests

### LINKED ISSUE #21: Write comprehensive parser tests for strings
This means:

- [ ] add tests to "literal string" section of ParserTest.elm about
  - [ ] multiline string (`"""abc"""`)
  - [ ] how do escapes work in normal vs multiline strings (`"{\"foo\":1}"` vs `"""{"foo":1}"""`)
  - [ ] that multiline strings actually parse newlines!
- [ ] update the "what's done" table in README - parser tests for strings

Some of the tests should fail - not implemented yet.
Fixing the parser is not needed in this ticket (can be in a separate one)

aaronjanse:
fix #21

The code for writing literals is, to say the least, messy.
I'm open to ideas for how to do this better, if possible.

### DIFF
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -585,7 +585,7 @@ expr =
                   , Ok (Literal (Char '\t'))
                   )
                 , ( "double quote"
-                  , singleQuote "\\\""
+                  , singleQuote "\""
                   , Ok (Literal (Char '"'))
                     -- " (for vscode-elm bug)
                   )
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -672,7 +672,7 @@ expr =
                   , Ok (Literal (String "'"))
                   )
                 , ( "double quote"
-                  , tripleQuote "\"\""
+                  , tripleQuote " \" "
                   , Ok (Literal (String "\""))
                   )
                 , ( "escape n"
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -406,17 +406,17 @@ moduleName =
 
 singleQuote : String -> String
 singleQuote txt =
-  "\'" ++ txt ++ "\'"
+    "'" ++ txt ++ "'"
 
 
 doubleQuote : String -> String
 doubleQuote txt =
-  "\"" ++ txt ++ "\""
+    "\"" ++ txt ++ "\""
 
 
 tripleQuote : String -> String
 tripleQuote txt =
-  "\"\"\"" ++ txt ++ "\"\"\""
+    "\"\"\"" ++ txt ++ "\"\"\""
 
 
 expr : Test
@@ -646,7 +646,7 @@ expr =
                   )
                 , ( "combo of escapes and chars"
                   , doubleQuote "\\u{1F648}\\n\\r\\t\\abc123"
-                  , Ok (Literal (String "🙈\n\r\t\\abc123"))
+                  , Ok (Literal (String "🙈\n\u{000D}\t\\abc123"))
                   )
                 ]
               )
@@ -697,7 +697,7 @@ expr =
                   )
                 , ( "combo of escapes, newlines, and chars"
                   , tripleQuote "\\u{1F648}\\n\n\n\\r\\t\\abc123"
-                  , Ok (Literal (String "🙈\n\n\n\r\t\\abc123"))
+                  , Ok (Literal (String "🙈\n\n\n\u{000D}\t\\abc123"))
                   )
                 ]
               )
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -590,7 +590,7 @@ expr =
                     -- " (for vscode-elm bug)
                   )
                 , ( "single quote"
-                  , singleQuote "'\\''"
+                  , singleQuote "\\'"
                   , Ok (Literal (Char '\''))
                   )
                 , ( "emoji"
@@ -646,7 +646,7 @@ expr =
                   )
                 , ( "combo of escapes and chars"
                   , doubleQuote "\\u{1F648}\\n\\r\\t\\abc123"
-                  , Ok (Literal (String "🙈\n\r\tabc123"))
+                  , Ok (Literal (String "🙈\n\r\t\\abc123"))
                   )
                 ]
               )
@@ -673,7 +673,7 @@ expr =
                   )
                 , ( "double quote"
                   , tripleQuote " \" "
-                  , Ok (Literal (String "\""))
+                  , Ok (Literal (String " \" "))
                   )
                 , ( "escape n"
                   , tripleQuote "\\n"
--- a/tests/ParserTest.elm
+++ b/tests/ParserTest.elm
@@ -635,6 +635,57 @@ expr =
                   )
                 ]
               )
+            , ( "literal multiline string"
+              , [ ( "empty"
+                  , "\"\"\"\"\"\""
+                  , Ok (Literal (String ""))
+                  )
+                , ( "one space"
+                  , "\"\"\" \"\"\""
+                  , Ok (Literal (String " "))
+                  )
+                , ( "newline"
+                  , "\"\"\"\n\"\"\""
+                  , Ok (Literal (String "\n"))
+                  )
+                , ( "two numbers"
+                  , "\"\"\"42\"\"\""
+                  , Ok (Literal (String "42"))
+                  )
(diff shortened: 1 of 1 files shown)


Janiczek:
yeah, we can't use the `""" ... """` trick I fear. This will have to do

tests/ParserTest.elm
```diff
@@ -605,6 +605,85 @@ expr =
...
+                  , "\"😃\""
+                  , Ok (Literal (String "😃"))
+                  )
+                , ( "escaped unicode code point"
+                  , "\"\\u{1F648}\""
+                  , Ok (Literal (String "🙈"))
+                  )
+                , ( "combo of escapes and chars"
+                  , "\"\\u{1F648}\\n\\r\\t\\abc123\""
+                  , Ok (Literal (String "🙈\n\r\tabc123"))
```
Janiczek: Is this `\r` consistent with the other test - that returns `\u{000D}`? If both would pass, why write the `\u{000D}` thing at all?

tests/ParserTest.elm
```diff
@@ -605,6 +605,85 @@ expr =
...
+                , ( "two numbers"
+                  , "\"\"\"42\"\"\""
+                  , Ok (Literal (String "42"))
+                  )
+                , ( "single quote"
+                  , "\"\"\"'\"\"\""
+                  , Ok (Literal (String "'"))
+                  )
+                , ( "double quote"
+                  , "\"\"\"\"\"\"\"\""
```
Janiczek: I worry this isn't a correct test - this doesn't work in official Elm REPL for example.
x

Janiczek: You can try to change it for something like `""" " """` - that demonstrates the inner `"` doesn't need escapes.

tests/ParserTest.elm
```diff
@@ -605,6 +605,85 @@ expr =
...
+                  , "\"\"\"😃\"\"\""
+                  , Ok (Literal (String "😃"))
+                  )
+                , ( "escaped unicode code point"
+                  , "\"\"\"\\u{1F648}\"\"\""
+                  , Ok (Literal (String "🙈"))
+                  )
+                , ( "combo of escapes, newlines, and chars"
+                  , "\"\"\"\\u{1F648}\\n\n\n\\r\\t\\abc123\"\"\""
+                  , Ok (Literal (String "🙈\n\n\n\r\tabc123"))
```
Janiczek: the slashes right before `abc123` aren't in the output string ... that's probably an error?

tests/ParserTest.elm
aaronjanse: `\r` and `\u{000D}` are exactly the same characters.

However, elm-format prefers that we use `\u{000D}` instead of `\r`. I assume that the Elm overlords are trying to deprecate `\r` or something, but, in the meantime, I assume we still need to know how to parse `\r` even if we never write it that way in our source code.

Unless you want to use `\r` everywhere, I should:
```diff
- , Ok (Literal (String "🙈\n\r\tabc123"))
+ , Ok (Literal (String "🙈\n\u{000D}\tabc123"))
```

Janiczek: OK, now that you mention it I remember `\r` being changed by `elm-format` under my hands. Let's keep it as you wrote it then.

tests/ParserTest.elm
aaronjanse: Interesting. `""""""""` works when running `make test` but not in the REPL.

However, `""" "" """` works in both

tests/ParserTest.elm
aaronjanse: It is an error. Fixed by https://github.com/elm-in-elm/compiler/pull/31/commits/b74c5be275ae43548dce26d2647ff165cebc2e70, thanks

aaronjanse:
> yeah, we can't use the `""" ... """` trick I fear. This will have to do

Could/should we make `tripleQuote` function?
```elm
tripleQuote txt = 
  "\"\"\"" ++ txt ++ "\"\"\""
```

tests/ParserTest.elm
Janiczek: `""""""""` probably works because there's an even number of `"` (8)
I'd expect `"""""""` (7) to not work in both

Janiczek:
> Could/should we make tripleQuote function?

That's a good idea - let's use that!

aaronjanse:
>> Could/should we make tripleQuote function?  That's a good idea - let's use that!
>
> That's a good idea - let's use that!

Should we create `doubleQuote` and `singleQuote` functions, too?

Janiczek:
Might look better? Let's try it, we can revert later

tests/ParserTest.elm
harrysarson: Maybe this would be easier to grok?

```suggestion
                  , String.repeat 7 "\""
```

Janiczek: Would be a good idea, but `"""""""` is not supposed to parse. I'm suggesting `""" " """`, at which point `tripleQuote """ " """` or `tripleQuote " \" "` is what we need

harrysarson: Ah I see! Fair point.

aaronjanse: Fixed by https://github.com/elm-in-elm/compiler/pull/31/commits/4bdfbef1a61dfa8c3448add3941f432c38ea895e

aaronjanse:
> Might look better? Let's try it, we can revert later

See https://github.com/elm-in-elm/compiler/pull/31/commits/65940dd97bbff9ba479cf61738885e72d0f7fd2e :+1:

tests/ParserTest.elm
```diff
@@ -558,53 +573,132 @@ expr =
...
                   )
                 , ( "double quote"
-                  , "'\\\"'"
+                  , singleQuote "\\\""
                   , Ok (Literal (Char '"'))
                     -- " (for vscode-elm bug)
                   )
                 , ( "single quote"
-                  , "'\\''"
+                  , singleQuote "'\\''"
```
Janiczek: Is this correct? I'd expect `singleQuote "'"`

tests/ParserTest.elm
```diff
@@ -558,53 +573,132 @@ expr =
...
+                  , tripleQuote "42"
+                  , Ok (Literal (String "42"))
+                  )
+                , ( "single quote"
+                  , tripleQuote "'"
+                  , Ok (Literal (String "'"))
+                  )
+                , ( "double quote"
+                  , tripleQuote " \" "
+                  , Ok (Literal (String "\""))
```
Janiczek: I'd expect the spaces to be in the parsed string too

aaronjanse: Fixed by https://github.com/elm-in-elm/compiler/pull/31/commits/51d0b7fbad7f45642e392142f5acad0f1d6ad1cc

tests/ParserTest.elm
aaronjanse: Fixed by https://github.com/elm-in-elm/compiler/pull/31/commits/51d0b7fbad7f45642e392142f5acad0f1d6ad1cc

tests/ParserTest.elm
```diff
@@ -558,53 +573,132 @@ expr =
...
                   , Ok (Literal (Char '\u{000D}'))
                   )
                 , ( "escape t"
-                  , "'\\t'"
+                  , singleQuote "\\t"
                   , Ok (Literal (Char '\t'))
                   )
                 , ( "double quote"
-                  , "'\\\"'"
+                  , singleQuote "\\\""
```
Janiczek: now that I look at it, I also wonder if this is right.

from the user perspective, `'"'`
from the parser input perspective, `"'\"'"`, which is `singleQuote "\""`
from the parser output perspective, `Char '"'`

WDYT?

Janiczek:
One last nitpick (I think :see_no_evil:), otherwise ready for merge! :+1:

tests/ParserTest.elm
aaronjanse: That's correct. Did I make a typo?

Janiczek: Yeah, if I see correctly you do `singleQuote "\\\""` instead of `singleQuote "\""`

aaronjanse:
Oh! That makes sense. I'll fix that this afternoon.

Or, if you want to, feel free to fix it yourself and merge :+1:

Janiczek:
@aaronjanse I'll do that. Thanks for all the effort that went into this!

PR #32, opened 2019-07-05
Implement the library

Janiczek:


Janiczek:
TODO: think about "pure" (no downloading packages, pure) + "practical" (downloading packages with Cmds) APIs

elm.json
```diff
@@ -1,49 +1,31 @@
 {
-    "type": "application",
-    "source-directories": [
-        "src/compiler",
-        "src/library",
-        "src/cli"
+    "type": "package",
```
Janiczek: Change from application to package, because for libraries the filepaths are kinda set in stone... top-level elm.json, src/ source directory...

src/Elm/Compiler.elm
```diff
@@ -0,0 +1,636 @@
+module Elm.Compiler exposing
```
Janiczek: The important file in this PR.

src/Stage/Emit.elm
```diff
@@ -0,0 +1,437 @@
+module Stage.Emit exposing
```
Janiczek: The Emit stage also got pretty refactored, to be more generic. The PrepareForBackend stage kicked the bucket, as we want to be generic and not model the whole compiler according to what makes sense in JS world.

src/todo.txt
```diff
@@ -0,0 +1,25 @@
+stuff to expose:
```
Janiczek: This TODO file is a list of ideas I still have to go through and decide for this to be an usable library.

harrysarson:
Fantastic!

Issue #17, opened 2019-06-30
Add floats to AST, write comprehensive parser tests for them


Janiczek:
This means:

- [ ] add Float constructor to the Literal type... solve any compiler errors that arise as a result (emitExpr etc - look at the Int cases for inspiration, it should be pretty similar)
- [x] find all the ways floats can be written in Elm (`1.2`? `1.`? scientific notation like `2e10`? maybe stuff like `NaN`, `Infinity` etc? (probably not))... go see official Elm's parser to be sure!
- [ ] add those as tests as a "literal float" section of ParserTest.elm (probably doesn't exist yet) so that the tests exercise all the possibilities and truthfully tell us what's done and what's not...
- [ ] add emit tests too
- [ ] update the "what's done" table in README - parser tests for floats, but probably also optimizations (none), desugaring, typechecking and emitting

Janiczek:
Seems like all the ways are:
* `1.2` (positive float)
* `-1.2` (negative float)
* `5.2e3` (scientific notation)
* `-5.2e3` (negative number before exponent)
* `5.2E3` (big E)
* `5e-2` (negative exponent)

aaronjanse:
Looks like adding it to the AST shouldn't be too difficult using [`Parser.float`](https://package.elm-lang.org/packages/elm/parser/latest/Parser#float) :+1:

Janiczek:
That might very well be the case, I'm just wary of the `elm/parser` bugs wrt. Parser.int and Parser.float:
https://github.com/elm/parser/issues

Anyways, even if we found out we can't use these functions and had to roll our own, doing what the reference implementation does shouldn't be _too_ hard: https://github.com/elm/compiler/blob/master/compiler/src/Parse/Number.hs

Warry:
The `number` parser seems adapted:

```
literalNumber : Parser_ Literal
literalNumber =
    let
        parseLitteralNumber =
            P.number
                { int = Ok Int
                , hex = Ok Int
                , octal = Ok Int
                , binary = Ok Int
                , float = Ok Float
                , invalid = InvalidNumber
                , expecting = ExpectingNumber 
                }

        negateLitteral toBeNegated =
            case toBeNegated of
                Int int ->
                    Int (negate int)

                Float float ->
                    Float (negate float)

                _ ->
                    toBeNegated

    in
    P.oneOf
        [ P.succeed negateLitteral
            |. P.symbol (P.Token "-" ExpectingMinusSign)
[...]

PR #36, opened 2019-07-08
Add floats and other numbers

Warry:


Janiczek:
fixes #17

Warry:
hum, I did not see that old commit sneaking in!

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -20,6 +20,9 @@ emitExpr ( expr, _ ) =
         Literal (Int int) ->
             String.fromInt int
 
+        Literal (Float float) ->
+            String.fromFloat float
```
Janiczek: Let's add some emit tests for this. We can try for a little bit more corner cases than with [Ints](https://github.com/elm-in-elm/compiler/blob/master/tests/EmitTest.elm#L42-L46) - eg. what does parsing `0.3` become? etc.

Then we'll be able to :heavy_check_mark: the emit tests in the README table.

Janiczek:
The diff in the PR is a bit weird now ... :thinking: maybe just some local cache. I'll try to deal with that.

PR #37, opened 2019-07-08
Add floats (with tests) fixes #17

### LINKED ISSUE #17: Add floats to AST, write comprehensive parser tests for them

Warry:


tests/EmitTest.elm
```diff
@@ -45,6 +45,15 @@ javascript =
                     , ( "negative int", Literal (Int -998), "-998" )
                     ]
                 )
+            , describe "Float"
+                (List.map runTest
+                    [ ( "positive float", Literal (Float 12.3), "12.3" )
+                    , ( "negative float", Literal (Float -12.3), "-12.3" )
+                    , ( "scientific notation", Literal (Float 1e2), "100" )
+                    , ( "negative scientific notation", Literal (Float -1e2), "-100" )
+                    , ( "negative exponent", Literal (Float 5e-2), "0.05" )
```
Janiczek: I don't think we need to concern ourselves with scientific notation (these last three tests) in emit tests.

If eg. `1e2` passes the parser, it goes into the rest of the pipeline as `100` ... so the test *doesn't* test anything about scientific notation, I believe. It is definitely useful as a parser test though! :+1:

README.md
```diff
@@ -71,7 +71,7 @@ Oh God please yes! :heart: Feel free to look around the [<kbd>help wanted</kbd>]
 |                   | parser tests         | optimize tests       | emit tests         | parse                | desugar            | infer types        | optimize             | emit               |
 | ----------------- | -------------------- | -------------------- | ------------------ | -------------------- | ------------------ | ------------------ | -------------------- | ------------------ |
 | integers          | :heavy_check_mark:   | :warning: [[2]](#f2) | :heavy_check_mark: | :heavy_check_mark:   | :heavy_check_mark: | :heavy_check_mark: | :warning: [[2]](#f2) | :heavy_check_mark: |
-| floats            | :x: [[3]](#f3)       | :x: [[5]](#f5)       | :x: [[3]](#f3)     | :x: [[4]](#f4)       | :x: [[3]](#f3)     | :x: [[3]](#f3)     | :x: [[5]](#f5)       | :x: [[3]](#f3)     |
+| floats            | :heavy_check_mark:   | :x: [[5]](#f5)       | :x: [[3]](#f3)     | :heavy_check_mark:   | :x: [[3]](#f3)     | :x: [[3]](#f3)     | :x: [[5]](#f5)       | :heavy_check_mark: |
```
Janiczek: emit tests can be :heavy_check_mark: instead of :x: here

tests/ParserTest.elm
```diff
@@ -553,6 +553,37 @@ expr =
...
+                  , "5.2E3"
+                  , Ok (Literal (Float 5.2e3))
+                  )
+                , ( "Negative scientific notation"
+                  , "-5.2e3"
+                  , Ok (Literal (Float -5.2e3))
+                  )
+                , ( "Negative exponent"
+                  , "5e-2"
+                  , Ok (Literal (Float 5e-2))
```
Janiczek: maybe let's make those "expected result" values normal float numbers, just to clearly see what the notation is doing, instead of relying on the Elm running the tests doing the right thing.

Can we change these to something like this?
```elm
                , ( "Negative exponent"
                  , "5e-2"
                  , Ok (Literal (Float 0.05)) ------------------------ change scientific notation in results to normal floats
                  )
```

tests/EmitTest.elm
```diff
@@ -45,6 +45,12 @@ javascript =
                     , ( "negative int", Literal (Int -998), "-998" )
                     ]
                 )
+            , describe "Float"
+                (List.map runTest
+                    [ ( "positive float", Literal (Float 12.3), "12.3" )
+                    , ( "negative float", Literal (Float -12.3), "-12.3" )
```
Janiczek: Can we please add a cornercase of `Float 0.0` and `Float -0.0`? Are they even different?
As we use `String.fromFloat` we're getting its behaviour, I'm not sure what it actually is :smile:

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -378,54 +378,47 @@ literal : Parser_ Frontend.Expr
...
-                            |> Result.map P.succeed
-                            |> Result.withDefault (P.problem InvalidHexInt)
-                    )
-                |> P.inContext InHexInt
+        parseLitteralNumber =
+            P.number
+                { int = Ok Int
+                , hex = Ok Int
+                , octal = Ok Int
+                , binary = Ok Int
```
Janiczek: Elm 0.19 doesn't support octal and binary numbers, so let's make these two rows `Nothing`

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -378,54 +378,47 @@ literal : Parser_ Frontend.Expr
...
-                    (\hexString ->
-                        hexString
-                            -- TODO this String.toLower shouldn't be needed - see https://github.com/rtfeldman/elm-hex/pull/1
-                            |> String.toLower
-                            |> Hex.fromString
-                            |> Result.map P.succeed
-                            |> Result.withDefault (P.problem InvalidHexInt)
-                    )
-                |> P.inContext InHexInt
+        parseLitteralNumber =
```
Janiczek: also, typo (`Litteral` :arrow_right: `Literal`)

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -378,54 +378,47 @@ literal : Parser_ Frontend.Expr
...
+                { int = Ok Int
+                , hex = Ok Int
+                , octal = Ok Int
+                , binary = Ok Int
+                , float = Ok Float
+                , invalid = InvalidNumber
+                , expecting = ExpectingNumber 
+                }
+
+        negateLitteral toBeNegated =
```
Janiczek: similar typo

tests/EmitTest.elm
```diff
@@ -45,6 +45,14 @@ javascript =
...
+                    , ( "positive zero float", Literal (Float 0.0), "0" )
+                    , ( "negative zero float", Literal (Float -0.0), "0" )
```
Janiczek: Thanks! Can we add this Ellie as a comment? I think it shows some of the behaviour pretty nicely. Eg.
* the positive and negative zeros are equal to each other
* but are different entities - when you divide by them, they result in positive or negative Infinity

Interesting to see that String.fromFloat shows them as the same thing!

Janiczek:
From running the tests it seems to me the `elm/parser` bug is now again present:
https://travis-ci.com/elm-in-elm/compiler/builds/118261553#L533

It's eating the `e` in `else`. The `backtrackable` call (https://github.com/elm-in-elm/compiler/pull/37/files#diff-27fa2f98c2b4d7c0696e05c001f6c1c6L394) was there to make it continue when it fails on that issue.

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -390,15 +390,16 @@ literalNumber : Parser_ Literal
...
-                , float = Ok Float
-                , invalid = InvalidNumber
-                , expecting = ExpectingNumber 
-                }
+            P.backtrackable <|
+                P.number
+                    { int = Ok Int
+                    , hex = Ok Int
+                    , octal = Err InvalidNumber -- Elm do not supports octal notation
+                    , binary = Err InvalidNumber -- Elm do not supports binary notation
```
Janiczek: just one last(:grey_question:) nitpick - English :smile: `Elm does not support ...`

README.md
Janiczek: This is what I meant:
emit

btw I think **we can :heavy_check_mark: everything except optimizations** in that column, and under the table make the `f3` and `f4` anchors similar to `f6` (recycle them: remove the note there but keep the footnote existing with something like `this space left intentionally blank` so that we don't have to cascade number changes into the table)

If my description of this is confusing (English is not my mother language either), tell me and I'll just do it myself :sweat_smile: #languageBarriers

Janiczek:
@Warry I'm happy with the tests and everything :+1: Pleasantly surprised we can use `Parser.numb
After the README changes I'll merge this. Thanks for all the work! :raised_hands:

PR #39, opened 2019-07-08
[wip] Add support for lists

### LINKED ISSUE #27: Add support for lists
This means:

- [ ] add `List (List Expr)` constructor `Frontend.Expr` type... solve any compiler errors that arise as a result (`emitExpr` etc)
- [ ] add tests that show all the various ways lists can be written work. For example:
  - [ ] `[]`, `[ ]`
  - [ ] `[1]`, `[ 1 ]`
  - [ ] `[1,2]`, `[ 1, 2 ]`
  - I can't think of any more. Let's not deal with newlines right now as that will be more tricky and done in another issue.
- [ ] partially implement type inference - focus on the items: all items inside list have to have the same type. Should be enough to eg. for list `[A,B,C]` do `equals typeA typeB, equals typeA typeC`. Don't have to do all the combinations. This will leave us to do the type of the actual list itself in a later issue after we implement type parameters.
- [ ] add emit tests too:
  - [ ] Elm's `[ 1, 2, 3 ]` :arrow_right: JavaScript's `[1, 2, 3]`
  - [ ] `[]` :arrow_right: `[]`
- [ ] implement emitting to pass the emit tests above
- [ ] update the "what's done" table in README for lists: parser tests, desugaring (none), type inference (only partially = :warning:), emit and emit tests... but not parsers and optimizations.

If you get stuck on anything, ask here or on Discord! We'll help you :slightly_smiling_face: 

## Out of scope :x: :x: :x:

- [ ] tests for newlines support in lists
- [ ] parser for lists
- [ ] optimizations (they'll become available with the operators but not sooner I think ... at least I can't think of any)
- [ ] type inference of list itself
- [ ] `(::)` operator
- [ ] `(++)` operator

Warry:
fixes #27


src/compiler/AST/Frontend.elm
```diff
@@ -31,6 +31,7 @@ type Expr
     | Call { fn : Expr, argument : Expr }
     | If { test : Expr, then_ : Expr, else_ : Expr }
     | Let { bindings : List (Binding Expr), body : Expr }
+    | List { list: List Expr }
```
harrysarson: What is the advantage of using a record here?

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -66,6 +66,11 @@ emitExpr ( expr, _ ) =
             in
             "((() => {" ++ bindingsJS ++ "; return " ++ emitExpr body ++ ";})())"
 
+        List { list } ->
+            "["
+            ++ (List.map emitExpr list |> String.join ", ")
+            ++ "]"
```
harrysarson: Official elm represents lists as linked lists rather than javascript arrays which makes concatenation work. I suspect that elm-in-elm should follow the same approach

harrysarson:
Looks nice!

src/compiler/AST/Frontend.elm
Janiczek: it might have been copying the style of constructors above.

my original intention with those was "if one argument, keep it naked, if it's eg. 3x Expr, name them to prevent confusion"

I'd unwrap the record in this case

Warry: I'm going to remove this code, it's out of the scope of this PR. It was more like an exercise.

Warry:
To clarify: I toyed around to get to understand what I'm touching. The given assignment doesn't cover parsing, so it'll be deleted.
My plan this evening:
- Seek what errors the elm compiler references
- Write code that cover more cases, including breaking
- Adding tests for the Type Inference stage

src/compiler/AST/Common/Type.elm
```diff
@@ -16,6 +16,7 @@ type Type
     | Char
     | String
     | Bool
+    | List (List Type)
```
Janiczek: I think it has to be `List Type` (like in real Elm it's not `[Int, Float, String]` but `List Int`)

src/compiler/Stage/Emit/JavaScript.elm
Janiczek: Interestingly official Elm keeps list literals as Lists of the language it's defined in (see [AST.Source](https://github.com/elm/compiler/blob/master/compiler/src/AST/Source.hs#L52), [AST.Canonical](https://github.com/elm/compiler/blob/master/compiler/src/AST/Canonical.hs#L91), [AST.Optimized](https://github.com/elm/compiler/blob/master/compiler/src/AST/Optimized.hs#L60)) and when emitting JS ([Generate.JavaScript.Expression](https://github.com/elm/compiler/blob/master/compiler/src/Generate/JavaScript/Expression.hs#L103-L113)) either returns `Nil` or calls `fromArray` and generates a JS list as an argument for it.

We'll definitely have to plug into the `elm/core` List definition (btw funny thing, it's [JS-only](https://github.com/elm/core/blob/1.0.2/src/Elm/Kernel/List.js) :sob: ), but we can do that later. What @Warry has right now is not that far off, and for now good enough I think. When we have enough stuff working that it stops being consistent with each other, we'll fix it.

WDYT @harrysarson @Warry ?

harrysarson: That is interesting, also interesting why the list definition is in javascript. 

Regarding long term plan sounds good - the point at which it stops working will probably be when we implement `(::)`.

Janiczek:
@Warry if you want to do everything in one go (ie. with parsers included), feel free to; I've just separated the issues as I think that would be more helpful (smaller problems) for newcomers. If you want to do parsers with the rest of the stuff, I won't mind.

src/compiler/Error.elm
```diff
@@ -116,6 +121,7 @@ type TypeError
     = UnknownName VarName
     | TypeMismatch Type Type
     | OccursCheckFailed Int Type
+    | ListTypeMismatch Int Type Type
```
Janiczek: Sorry if I'm commenting on what you'll want to change later! Dunno which parts are still WIP :)

I think this error has a bit better knowledge of the types than the generic `TypeMismatch`.
This one can say "the nth type was Foo, all before were Bar!"... I'd make the parameters a record
```elm
| ListTypeMismatch
    { position : Int
    , type_ : Type
    , previousTypes : Type
    }
```
to prevent the confusion between which `Type` is which, and edit the message to say something like "This element has type Foo, and all the previous elements had type Bar".

src/compiler/Stage/InferTypes.elm
```diff
@@ -251,6 +251,19 @@ generateEquations ( expr, type_ ) =
                 :: generateEquations body
                 ++ bindingEquations
 
+        Typed.List list ->
+            case type_ of
+                Type.List type1 ->
+                    List.foldl
+                        (\(entry, type2) acc ->
+                            (equals type1 type2) :: acc
```
Janiczek: I'd just change the names a bit:
* `type1` is `listParamType`
* `entry` we don't care about :arrow_right: `_`
* `type2` can be `listElementType`
* and `acc` is `equations` or `eqs`?

Anyway that's how I'd write it. :sweat_smile:

tests/EmitTest.elm
```diff
@@ -163,6 +163,22 @@ javascript =
                       )
                     ]
                 )
+            , describe "List"
+                (List.map runTest
```
Janiczek: Let's write a comment here that this `[...]` format will eventually have to change to something like `List$fromArray([...])` when we add `(::)`, `(++)`, etc - this piggybacking on platform's collections has to end somewhere and get converted to the actual Elm values

tests/InferTypesTest.elm
```diff
@@ -0,0 +1,64 @@
...
+                Expect.fail <|
+                    case typeError of
+                        Error.TypeMismatch t1 t2 ->
+                            "The types "
+                                ++ Type.toString t1
+                                ++ " and "
+                                ++ Type.toString t2
+                                ++ " don't match."
+                        _ ->
+                            "wft"
```
Janiczek: probably just use `Error.toString`? :see_no_evil:

Warry: yes! The state of that PR is a mess, I won't leave it that way very long ^^ I guess I just have to `Error.toString Error.TypeError typeError`.

tests/ParserTest.elm
```diff
@@ -764,6 +764,33 @@ expr =
...
+                  , "[ 1 ]"
+                  , Ok (AST.Frontend.List [ Literal (Int 1) ])
+                  )
+                , ( "simple list"
+                  , "[1,2,3]"
+                  , Ok (AST.Frontend.List [ Literal (Int 1), Literal (Int 2), Literal (Int 3) ])
+                  )
+                , ( "simple list with inner spaces"
+                  , "[ 1,  2  , 3 ]"
+                  , Ok (AST.Frontend.List [ Literal (Int 1), Literal (Int 2), Literal (Int 3) ])
```
Janiczek: I think we can drop the full qualification (`AST.Frontend.???`) ... can you do this for both List and Let tests please?

tests/InferTypesTest.elm
```diff
@@ -0,0 +1,59 @@
...
+    describe "Stage.InferType"
+        (List.map runSection
+            [ ( "list"
+              , [ ( "different types"
+                  , Canonical.List [ Canonical.Literal (Int 1), Canonical.Literal (String "two") ]
+                  , Err (Error.TypeMismatch Type.Int Type.String)
+                  )
+                , ( "more items types"
+                  , Canonical.List [ Canonical.Literal (Int 2), Canonical.Literal (Int 3), Canonical.Literal (Bool False) ]
+                  , Err (Error.TypeMismatch (Type.Int) (Type.Bool))
```
Janiczek: some extra parentheses, I expect running elm-format would solve those

tests/InferTypesTest.elm
```diff
@@ -0,0 +1,59 @@
...
+
+        runTest ( description, input, output ) =
+            test description <|
+                \() ->
+                    Stage.InferTypes.inferExpr input
+                        |> expectEqualInferType output
+    in
+    describe "Stage.InferType"
+        (List.map runSection
+            [ ( "list"
```
Janiczek: can we run some tests that expect success?

tests/InferTypesTest.elm
```diff
@@ -0,0 +1,59 @@
...
+              )
+            ]
+        )
+
+
+expectEqualInferType :
+    Result TypeError Typed.Expr
+    -> Result TypeError Typed.Expr
+    -> Expectation
+expectEqualInferType expected actual =
```
Janiczek: This function is a bit puzzling to me. Why does it automatically fail if `actual` is `Err`? What if `expected` is `Err` too?

tests/EmitTest.elm
```diff
@@ -163,6 +162,28 @@ javascript =
...
+                      , List [ typed (Literal (Int 1)) ]
+                      , "[1]"
+                      )
+                    , ( "simple list"
+                      , List [ typed (Literal (Int 1)), typed (Literal (Int 2)), typed (Literal (Int 3)) ]
+                      , "[1, 2, 3]"
+                      )
+                    , ( "simple list2"
+                      , List [ typed (Literal (String "one")), typed (Literal (Int 2)), typed (Literal (Int 3)) ]
+                      , "[\"one\", 2, 3]"
```
Janiczek: I'd drop this test as that source Expr is a bit nonsensical. Or let's change it to a list of three strings?

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -660,6 +661,20 @@ unit _ =
...
+    P.succeed Frontend.List
+        |= P.sequence
+            { start = P.Token "[" ExpectingLeftBracket
+            , separator = P.Token "," ExpectingListSeparator
+            , end = P.Token "]" ExpectingRightBracket
+            , spaces = spacesOnly
+            , item = PP.subExpression 0 config
+            , trailing = P.Forbidden
+            }
+        |> P.inContext InList
```
Janiczek: Cool to see `P.sequence` works well. I'm a bit careful regarding those helpers :sweat_smile:

src/compiler/Error.elm
```diff
@@ -202,6 +207,22 @@ toString error =
                     "Couldn't find the value `main` in the main module given to the compiler!"
 
 
+ordinal : Int -> String
+ordinal position =
```
Janiczek: this function is unused, can be deleted

src/compiler/Error.elm
```diff
@@ -99,6 +103,7 @@ type ParseProblem
     | ExpectingFalse
     | ExpectingLet
     | ExpectingIn
+    | ExpectingList
```
Janiczek: this is probably unused - can't find it anywhere

src/compiler/Stage/Desugar.elm
```diff
@@ -63,6 +63,9 @@ desugarExpr modules thisModule expr =
         Frontend.Let { bindings, body } ->
             desugarLet recurse bindings body
 
+        Frontend.List list_ ->
+            desugarList recurse list_
```
Janiczek: unneeded `_` in the `list_` variable? `list` or `items` will work

src/compiler/Stage/Desugar.elm
```diff
@@ -146,6 +149,13 @@ desugarLet recurse bindings body =
         (recurse body)
 
 
+desugarList : (Frontend.Expr -> Result DesugarError Canonical.Expr) -> List Frontend.Expr -> Result DesugarError Canonical.Expr
+desugarList recurse list_ =
+    List.map recurse list_
+        |> List.foldr (Result.map2 (::)) (Ok [])
+        |> Result.map Canonical.List
```
Janiczek: Again `list_` - can be `list` or `items`

Just to check my understanding - `foldr` is here to not reverse the list accidentally?

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -66,6 +66,11 @@ emitExpr ( expr, _ ) =
             in
             "((() => {" ++ bindingsJS ++ "; return " ++ emitExpr body ++ ";})())"
 
+        List list_ ->
```
Janiczek: again `list_` -> `items` or `list` :smile:

src/compiler/Stage/InferTypes/AssignIds.elm
```diff
@@ -170,3 +170,21 @@ assignIdsWith idSource expr =
 
         Canonical.Unit ->
             wrap idSource Typed.Unit
+
+        Canonical.List list_ ->
```
Janiczek: aand again `list_` -> `items` or `list`. Seeing as the rest of the function uses `items_`, I guess `items` makes most sense here.

src/compiler/Stage/Parse/Parser.elm
Warry: It's not perfect, but I guess it was specifically design with that usecase in mind, so I'm confident it can stay like this until perfs become a thing.

src/compiler/Stage/Desugar.elm
```diff
@@ -146,6 +149,13 @@ desugarLet recurse bindings body =
         (recurse body)
 
 
+desugarList : (Frontend.Expr -> Result DesugarError Canonical.Expr) -> List Frontend.Expr -> Result DesugarError Canonical.Expr
+desugarList recurse list =
+    List.map recurse list
+        |> List.foldl (Result.map2 (::)) (Ok [])
```
Janiczek: some `Debug.log`s:
```
source:           [ 1, 50, 999 ]
after parsing:    List [Literal (Int 1),Literal (Int 50),Literal (Int 999)]
after desugaring: List [Literal (Int 999),Literal (Int 50),Literal (Int 1)]
after typing:     (List [(Literal (Int 999),Int),(Literal (Int 50),Int),(Literal (Int 1),Int)],Var 4)
after emit:       [999, 50, 1]
```

I believe the `foldr` was correct. The comment below was just a question of why is the `foldr` there, didn't want to indicate that it's wrong ... :) sorry for confusion!

PR #43, opened 2019-07-11
Type to string

### LINKED ISSUE #41: Improve type variables in `AST.Common.Type.toString`
Right now type variables get emitted this way:
* `Type.Var 123` -> `"t123"`

It would be great to print them like `a` and `b`, like we're used to from the official Elm compiler.

### Gotchas

Note that `toString` for one `Type` is not enough. In the error messages you sometimes compare two types together, so something like `toStringHelper : Dict Int String -> Type -> (String, Dict Int String)` might be needed to allow you to thread names from one type into the other.

Eg. if the first Type is `Function (Var 1) (Function (Var 2) Int)` and the second one is `Function (Var 2) Bool`, they should be consistent with each other: `a -> b -> Int` and `b -> Bool` (not `a -> Bool`, which would probably happen if you'd just call `toString` without context and without threading the Dict through)

Warry:
fixes #41
also add parens when needed:
`List ((a -> b) -> b)`

### DIFF
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -31,10 +31,11 @@ getVarId type_ =
         _ ->
             Nothing
 
+
 toString : Type -> String
 toString type_ =
     getTypeVariablesIndex type_
-    |> toStringHelp type_
+        |> toStringHelp type_
 
 
 toStringHelp : Type -> Dict Int String -> String
@@ -42,7 +43,7 @@ toStringHelp type_ dict =
     case type_ of
         Var int ->
             Dict.get int dict
-            |> Maybe.withDefault ("t" ++ String.fromInt int)
+                |> Maybe.withDefault ("t" ++ String.fromInt int)
 
         Function t1 t2 ->
             wrapParens t1 dict ++ " -> " ++ toStringHelp t2 dict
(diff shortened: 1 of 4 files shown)


tests/InferTypesTest.elm
```diff
@@ -52,3 +53,56 @@ typeInference =
...
+                    Type.toString input
+                        |> Expect.equal output
+    in
+    describe "Type.toString"
+        (List.map runSection
+            [ ( "list"
+              , [ ( "empty list"
+                  , Type.List (Type.Var 0)
+                  , "List a"
+                  )
```
Janiczek: Can we make a test that running it with `Type.List (Type.Var 1)` instead of `0` also results in `List a`?

src/compiler/AST/Common/Type.elm
```diff
@@ -55,7 +63,71 @@ toString type_ =
...
+        Bool ->
+            (freeLetters, dict)
+
+        Unit ->
+            (freeLetters, dict)
+
+
+-- Stops at letter `s`, so that outbound variables display `tN`
+typeVariablesLetters =
+    [ "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s" ]
```
Janiczek: Interestingly `elm/compiler` does this:
```hs
fromTypeVariableScheme :: Int -> Name
fromTypeVariableScheme scheme =
  runST
  (
    if scheme < 26 then
      do  mba <- newByteArray 1
          writeWord8 mba 0 (0x61 + fromIntegral scheme)
          freeze mba
    else
      do  let (extra, letter) = quotRem scheme 26
          let !size = 1 + getIndexSize extra
          mba <- newByteArray size
          writeWord8 mba 0 (0x61 + fromIntegral letter)
          writeDigitsAtEnd mba size extra
          freeze mba
  )
```
which, I think, roughly translates to `a...z`, then `a1...z1`, then `a2...z2`, etc. into infinity.
**I think what we do is enough** - I mean, if you do so many type variables, something is wrong ... :)

src/compiler/AST/Common/Type.elm
```diff
@@ -55,7 +63,71 @@ toString type_ =
...
+
+                hd :: xs ->
+                    if Dict.member int dict then
+                        (freeLetters, dict)
+                    else
+                        (xs, Dict.insert int hd dict)
+
+        Function t1 t2 ->
+            getTypeVariablesIndexHelp t1 (freeLetters, dict)
+            |> getTypeVariablesIndexHelp t2
```
Janiczek: this is totally a nitpick but I think writing it as
```elm
(freeLetters, dict)
  |> getTypeVariablesIndexHelp t1
  |> getTypeVariablesIndexHelp t2
```
would better show the symmetry - that it recurses for each of its child expressions.

src/compiler/AST/Common/Type.elm
Warry: yeah, not sure there is anything to gain here... :)

Janiczek:
I think this API doesn't allow for the usecase of, say, writing out the error
```elm
TypeMismatch
  (List (Function (Var 0) (Var 1))
  (List (Function (Var 1) (Var 2))
```
such that it would print the types as `List (a -> b)` and `List (b -> c)` respectively.

Ie. there's no way to reuse the `( List String, Dict Int String )` state - `getTypeVariablesIndex` throws a part of the state away and only gives you the `Dict`.

What about:
```elm
type State = -- exposed *opaquely*
  State
    { freeLetters : List String
    , mapping : Dict Int String -- I find `index` confusing - makes me think it's a number instead of a Dict
    }
empty : State -- exposed
empty =
  State
    { freeLetters = [...]
    , mapping = Dict.empty
    }

toString : State -> Type -> (String, State)
```
such that it at least nudges the use that he'll probably have to thread the state through if he has multiple `toString` calls?

Warry:
well, it's possible to use `toStringHelp` to share the (freeLetters, dict) from one to another.

src/compiler/AST/Common/Type.elm
turboMaCk: I disagree. You're likely to run into problems with this sooner than you realize. Solution to this is very simple. Just have a function `Int -> Name` as elm does and do module of int and lenght of `typeVariablesLetters`. I would also suggest to define this as Array rather than List. I don't think so small simplification is worth the potential trouble.

Janiczek: @turboMaCk What's the usecase where you'd run into more than ~15 of those? IMHO it would have to be some monstrous `(a -> b -> c -> d -> ... -> z)` map26 thing. This doesn't concern situations like "there are more than 15 type variables in the whole module / project" - this is just for when printing types.

Although a generic function `Int -> String` would have its beauty - you'd be able to get rid of `freeLetters` entirely and only pass the Dict around :thinking:

turboMaCk: say person just don't use type annotation in whole program. Also do you already know how you'll be dealing with scopes. Elm don't have "ScopedTypeVariables". What about function with large let block. don't forget `a` on top is not the same as `a` in function in let block and that is not same `a` as a `a` in another function in the let block. You probably don't want to show error message like `expected a but was given a` right?

Janiczek: I'm fairly certain errors like `expected a but was given a` are orthogonal to the `Int -> String` stopping at `s` and then continuing with `t19`, `t20`, `t21`, etc. into infinity.

Anyway without an example that with some Elm coding style you can comfortably get into, say, `g` and above, the biggest motivation for "infinite" `Int -> String` function so far is just the simplicity of it.

turboMaCk: now I see the function https://github.com/elm-in-elm/compiler/pull/43/files#diff-8483228d929c9cb45d74eeaef0d7cc84R46 sorry I missunderstood first comment. So type var name collision can't happen. Though I would still say that Array would be more logical store to store these. Or maybe even Dict given ids are string values.

src/compiler/AST/Common/Type.elm
```diff
@@ -30,32 +42,103 @@ getVarId type_ =
...
 
         Unit ->
-            "()"
+            ( "Unit", state )
+
+
+dump : Type -> String
+dump type_ =
+    toString type_ empty
+        |> Tuple.first
```
Janiczek: I'd either remove this or add a comment along the lines of "Don't use this for any serious work - use `toString` instead. For debugging only!"

PR #44, opened 2019-07-11
Fuzz type inference

### LINKED ISSUE #40: Fuzzing type inference with expressions that type
This means:

* [ ] add an `InferTypesTest.elm` in the test project. (You can skip that if it's already done. #39 might do that)
* [ ] add a `typedExprFuzzer : Type -> Fuzzer Expr` function. This should generate a random expression that conforms to the type.
* [ ] add the above to write a `fuzzInference : Type -> Test` function. This should use `typedExprFuzzer` to test that the random expressions have the input type inferred.
* [ ] add type inference tests of the form `fuzzInference someTypeToCheck`. You can put them in a nested `describe` for grouping.

The idea is that generating expressions that fit a type should be easier than inferring the type of an expression.

If you get stuck on anything, ask here or on Discord! We'll help you :slightly_smiling_face:

## Out of scope :x: :x: :x:

* [ ] Error cases: we are not testing when inference fails and what it reports yet.
* [ ] Configuring the depth of the fuzzed expressions.

szabba:
fixes #40

### DIFF
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -174,12 +174,10 @@ letter int =
 
     maybeWrapParens (List Int) ( "List Int", state )
 
-
     -->
     ( "(List Int)", state )
 
     maybeWrapParens Int ( "Int", state )
-
 
     -->
     ( "Int", state )
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -24,6 +24,8 @@ type Type
     | Bool
     | List Type
     | Unit
+    | Tuple Type Type
+    | Tuple3 Type Type Type
 
 
 getVarId : Type -> Maybe Int
@@ -120,6 +122,29 @@ toString state type_ =
 
         Unit ->
             ( "Unit", state )
+
+        Tuple t1 t2 ->
+            let
+                ( t1String, state1 ) =
+                    toString state t1
+
+                ( t2String, state2 ) =
+                    toString state1 t2
+            in
+            ( "( " ++ t1String ++ ", " ++ t2String ++ " )", state2 )
+
+        Tuple3 t1 t2 t3 ->
+            let
+                ( t1String, state1 ) =
+                    toString state t1
+
+                ( t2String, state2 ) =
+                    toString state1 t2
+
+                ( t3String, state3 ) =
+                    toString state2 t3
+            in
+            ( "( " ++ t1String ++ ", " ++ t2String ++ ", " ++ t3String ++ " )", state3 )
 
 
 getName : State -> Int -> ( String, State )
@@ -225,4 +250,10 @@ shouldWrapParens type_ =
             True
 
         Unit ->
+            False
+
+        Tuple _ _ ->
+            False
+
+        Tuple3 _ _ _ ->
             False
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -1,6 +1,7 @@
 module AST.Common.Type exposing
     ( Type(..)
     , getVarId
+    , isNotParametric
     , toString
     )
 
@@ -28,6 +29,23 @@ getVarId type_ =
 
         _ ->
             Nothing
+
+
+isNotParametric : Type -> Bool
+isNotParametric type_ =
+    case type_ of
+        Var _ ->
+            False
+
+        Function input output ->
+            [ input, output ]
+                |> List.all isNotParametric
+
+        List element ->
+            element |> isNotParametric
+
+        _ ->
+            True
 
 
 toString : Type -> String
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -1,12 +1,17 @@
 module AST.Common.Type exposing
-    ( Type(..)
+    ( State
+    , Type(..)
+    , emptyState
     , getVarId
     , isNotParametric
+    , niceVarName
     , toString
     )
 
 {-| In its own module because both Error.TypeError and AST.Typed need to see it
 -}
+
+import Dict exposing (Dict)
 
 
 type Type
@@ -48,32 +53,178 @@ isNotParametric type_ =
             True
 
 
-toString : Type -> String
-toString type_ =
+
+--------------
+-- toString --
+--------------
+
+
+type State
+    = State
+        { mapping : Dict Int String
+        , counter : Int
+        }
+
+
+emptyState : State
+emptyState =
+    State
+        { mapping = Dict.empty
+        , counter = 0
+        }
+
+
+toString : State -> Type -> ( String, State )
+toString state type_ =
     case type_ of
         Var int ->
-            "t" ++ String.fromInt int
+            getName state int
 
         Function t1 t2 ->
-            toString t1 ++ " -> " ++ toString t2
+            let
+                ( t1String, state1 ) =
+                    toString state t1
+                        |> maybeWrapParens t1
 
-        Int ->
-            "Int"
-
-        Float ->
-            "Float"
-
-        Char ->
-            "Char"
-
-        String ->
-            "String"
-
-        Bool ->
-            "Bool"
+                ( t2String, state2 ) =
+                    toString state1 t2
+            in
+            ( t1String ++ " -> " ++ t2String
+            , state2
+            )
 
         List param ->
-            "List " ++ toString param
+            let
+                ( paramString, state1 ) =
+                    toString state param
+                        |> maybeWrapParens param
+            in
+            ( "List " ++ paramString
+            , state1
+            )
+
+        Int ->
+            ( "Int", state )
+
+        Float ->
+            ( "Float", state )
+
+        Char ->
+            ( "Char", state )
+
+        String ->
+            ( "String", state )
+
+        Bool ->
+            ( "Bool", state )
 
         Unit ->
-            "()"
+            ( "Unit", state )
+
+
+getName : State -> Int -> ( String, State )
+getName ((State { counter, mapping }) as state) varId =
+    case Dict.get varId mapping of
+        Just letter_ ->
+            ( letter_, state )
+
+        Nothing ->
+            let
+                name =
+                    niceVarName counter
+            in
+            ( name
+            , State
+                { counter = counter + 1
+                , mapping = Dict.insert varId name mapping
+                }
+            )
+
+
+{-| We're generating the nice variable names using this sequence:
+
+    a, b, ..., z, | a1, b1, ..., z1, | a2, b2, ..., z2, | ...
+    0  1       25 | 26  27       51  | 52  53       77  |
+
+-}
+niceVarName : Int -> String
+niceVarName id =
+    if id < 26 then
+        letter id
+
+    else
+        let
+            wraparounds =
+                id // 26
+
+            remainder =
+                id |> modBy 26
+        in
+        letter remainder ++ String.fromInt wraparounds
+
+
+letter : Int -> String
+letter int =
+    (int + 97 {- `a` -})
+        |> Char.fromCode
+        |> String.fromChar
+
+
+{-|
+
+    maybeWrapParens (List Int) ( "List Int", state )
+
+
+    -->
+    ( "(List Int)", state )
+
+    maybeWrapParens Int ( "Int", state )
+
+
+    -->
+    ( "Int", state )
+
+-}
+maybeWrapParens : Type -> ( String, a ) -> ( String, a )
+maybeWrapParens type_ ( string, state ) =
+    let
+        wrapParens : String -> String
+        wrapParens x =
+            "(" ++ x ++ ")"
+    in
+    if shouldWrapParens type_ then
+        ( wrapParens string, state )
+
+    else
+        ( string, state )
+
+
+shouldWrapParens : Type -> Bool
+shouldWrapParens type_ =
+    case type_ of
+        Var _ ->
+            False
+
+        Function t1 t2 ->
+            True
+
+        Int ->
+            False
+
+        Float ->
+            False
+
+        Char ->
+            False
+
+        String ->
+            False
+
+        Bool ->
+            False
+
+        List param ->
+            True
+
+        Unit ->
+            False
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -3,7 +3,7 @@ module AST.Common.Type exposing
     , Type(..)
     , emptyState
     , getVarId
-    , isNotParametric
+    , isParametric
     , niceVarName
     , toString
     )
@@ -38,21 +38,21 @@ getVarId type_ =
             Nothing
 
 
-isNotParametric : Type -> Bool
-isNotParametric type_ =
+isParametric : Type -> Bool
+isParametric type_ =
     case type_ of
         Var _ ->
-            False
+            True
 
         Function input output ->
             [ input, output ]
-                |> List.all isNotParametric
+                |> List.any isParametric
 
         List element ->
-            element |> isNotParametric
+            isParametric element
 
         _ ->
-            True
+            False
 
 
 
--- a/src/compiler/AST/Common/Type.elm
+++ b/src/compiler/AST/Common/Type.elm
@@ -51,6 +51,14 @@ isParametric type_ =
         List element ->
             isParametric element
 
+        Tuple left right ->
+            [ left, right ]
+                |> List.any isParametric
+
+        Tuple3 left middle right ->
+            [ left, middle, right ]
+                |> List.any isParametric
+
         _ ->
             False
 
--- a/tests/InferTypesFuzz.elm
+++ b/tests/InferTypesFuzz.elm
@@ -187,6 +187,8 @@ dumpType type_ =
         |> Tuple.first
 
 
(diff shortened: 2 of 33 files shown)


tests/InferTypesTest.elm
```diff
@@ -8,47 +8,142 @@ import Common
...
+        |> Fuzz.map Char
+        |> Fuzz.map Canonical.Literal
+
+
+stringExpr : Fuzzer Canonical.Expr
+stringExpr =
+    Fuzz.intRange 0 0x0010FFFF
+        |> Fuzz.list
+        |> Fuzz.map (List.map Char.fromCode)
+        |> Fuzz.map String.fromList
```
Janiczek: That's pretty bold :smile: but I guess for this usecase it doesn't matter if the strings are valid Unicode or not.

src/compiler/AST/Common/Type.elm
```diff
@@ -156,11 +174,13 @@ letter int =
 
     maybeWrapParens (List Int) ( "List Int", state )
 
+
```
Janiczek: for some reason these start piling up :sweat_smile: ideally there would be none between `maybeWrapParens` and `( "List Int)", state)`

src/compiler/AST/Common/Type.elm
```diff
@@ -3,6 +3,7 @@ module AST.Common.Type exposing
     , Type(..)
     , emptyState
     , getVarId
+    , isNotParametric
```
Janiczek: `isNotParametric` seems pretty specific for being an exposed function here :) Can we do `isParametric` and then `not` at the usage site?

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+exprGenerator : Type -> Generator Canonical.LocatedExpr
+exprGenerator targetType =
+    let
+        cannotFuzz details =
+            let
+                prefix =
+                    "Cannot fuzz `" ++ dumpType targetType ++ "` expressions."
+
+                message =
+                    if details |> String.isEmpty then
```
Janiczek: unnecessary `|>` here and there (see lines 37, 59, 145, 193, 267)

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+    type_
+        |> Type.toString Type.emptyState
+        |> Tuple.first
+
+
+exprShrinker : Shrinker Canonical.LocatedExpr
+exprShrinker expr =
+    case
+        expr |> Located.unwrap
+    of
```
Janiczek: this can be written on one line (I'm nitpicking, I know...)

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+        Canonical.Literal lit ->
+            lit
+                |> shrinkLiteral
+                |> Shrink.map Canonical.Literal
+                |> Shrink.map located
+
+        Canonical.Plus left right ->
+            expr
+                |> Debug.log "shrinking (+)"
```
Janiczek: Is this still needed?

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+        Bool b ->
+            b |> Shrink.bool |> Shrink.map Bool
+
+        String s ->
+            s |> Shrink.string |> Shrink.map String
+
+
+shrinkPlus : Canonical.LocatedExpr -> Canonical.LocatedExpr -> Shrinker Canonical.LocatedExpr
+shrinkPlus left right _ =
+    -- TODO: The Shrink docs were misleading here. Consider reporting an issue.
```
Janiczek: Care to elaborate more in the comment (or report the issue right away) while you still have the context in your head? This comment as it is now, doesn't give me enough information to understand what is the problem

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+
+
+shrinkNonEmptyList : Shrinker a -> Shrinker (List a)
+shrinkNonEmptyList shrinkElement list =
+    case list of
+        [] ->
+            list |> Shrink.noShrink
+
+        first :: rest ->
+            -- TODO: The Shrink docs were misleading here. Consider reporting an issue.
```
Janiczek: Same here as in `shrinkPlus`

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+                |> shrinkNonEmptyList exprShrinker
+                |> Shrink.map Canonical.List
+                |> Shrink.map located
+
+        Canonical.If { then_, else_ } ->
+            [ then_, else_ ]
+                |> List.map shrinkTo
+                |> concatShrink
+                -- Lesser hack.
+                |> shrink expr
```
Janiczek: So.. the `shrink` function exists for readability reasons? So that this isn't something like
```elm
            ([ then_, else_ ]
                |> List.map shrinkTo
                |> concatShrink
            ) expr
```
?

src/compiler/AST/Common/Type.elm
szabba: I wrote it as `isConcrete` first but then worried about that not being clear enough / aligned with the literature terminology. `isParametric` should be better.

tests/InferTypesFuzz.elm
szabba: I'm still doing some work on the PR. The comment is no longer present locally.

The issue is reported at https://github.com/elm-explorations/test/issues/101.

tests/InferTypesFuzz.elm
szabba: Exactly. It's not a great solution. I am very open to better ones.

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,293 @@
...
+                |> Debug.log "shrinking (+)"
+                |> shrinkPlus left right
+
+        Canonical.List elements ->
+            elements
+                |> shrinkNonEmptyList exprShrinker
+                |> Shrink.map Canonical.List
+                |> Shrink.map located
+
+        Canonical.If { then_, else_ } ->
```
Janiczek: no `test`?

szabba: Added a TODO about that.

tests/InferTypesTest.elm
```diff
@@ -231,3 +276,52 @@ niceVarName =
...
+            , ( Tuple3 String (Function (Var 0) Int) Unit, True )
+            , ( Tuple3 Bool Unit (Tuple (Var 0) Int), True )
+            ]
+
+
+dumpType : Type -> String
+dumpType type_ =
+    type_
+        |> Type.toString Type.emptyState
+        |> Tuple.first
```
Janiczek: There is `dumpType` in both InferTypesFuzz and InferTypesTest - we can move it to TestHelpers

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+import Common.Types exposing (VarName(..))
+import Fuzz exposing (Fuzzer)
+import Random exposing (Generator)
+import Random.Extra as Random
+import Shrink exposing (Shrinker)
+import TestHelpers exposing (located)
+
+
+exprTyped : Type -> Fuzzer Canonical.LocatedExpr
+exprTyped targetType =
```
Janiczek: I wonder about this name. Maybe `canonicalExpr` or `canonicalExprOfType`? If we don't plan to fuzz other expressions (Frontend / Typed), `exprOfType` would probably be the best. WDYT?

Janiczek: also it's weird that this module is named InferTypesFuzz when what it exposes is only a `Canonical.Expr` fuzzer :thinking:

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+
+        Type.String ->
+            stringExpr
+
+        Type.Unit ->
+            unitExpr
+
+        Type.List elementType ->
+            if Type.isParametric elementType then
+                cannotFuzz "Only lists with non-parametric element types are supported."
```
Janiczek: we could probably return an empty list here?

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+    case Located.unwrap expr of
+        Canonical.Literal lit ->
+            lit
+                |> shrinkLiteral
+                |> Shrink.map Canonical.Literal
+                |> Shrink.map located
+
+        Canonical.Plus left right ->
+            expr
+                |> shrinkPlus left right
```
Janiczek: The Shrinker API is really weird. I'd rather have the right amount of arguments and not give it useless arguments, bearing the cost of not being able to have a type annotation...
`shrinkPlus left right`
```elm
{-| shrinkPlus : Canonical.LocatedExpr -> Canonical.LocatedExpr -> LazyList Canonical.LocatedExpr
-}
shrinkPlus left right =
  lazyMap2 Canonical.Plus
    (exprShrinker left)
    (exprShrinker right)
    |> Shrink.map located
```
Would that work? (I haven't tried to run it)

Janiczek: It's nice to see that a new version of the API will hide LazyList totally and allow you to `Simplify.fromFunction : (a -> List a) -> Simplifier a`. Can't wait!
https://github.com/elm-explorations/test/blob/master/src/Simplify.elm

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+
+        Canonical.List elements ->
+            elements
+                |> shrinkNonEmptyList exprShrinker
```
Janiczek: Just checking my understanding... we're not shrinking `[1]` to `[]` because that would make the type not `List Int` but `List a`, which in itself isn't a problem (`List a` can be used where you need `List Int`) but our current testing solution doesn't currently have good functions to check that? (ie. we only have equality, which will say `List Int /= List a`)

ie. shrinking to empty list would have to be accompanied with a different operator in the test cases themselves (typechecks / is a subtype / unifies / ... instead of equality)

:question:

tests/InferTypesFuzz.elm
Janiczek: I think I'd rather have the code in the snippet in my original comment, and some comment explaining that the API is weird... and remove the `shrink` function. It is kinda just an additional indirection.

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+        Canonical.If { then_, else_ } ->
+            -- TODO: Shrink the test once we get more interesting Bool expressions.
+            [ then_, else_ ]
+                |> List.map shrinkTo
+                |> concatShrink
+                |> shrink expr
+
+        Canonical.Lambda { argument, body } ->
+            expr
+                |> shrinkLambda argument body
```
Janiczek: Let's do here the same thing as with the Plus.

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+
+        _ ->
+            expr |> Shrink.noShrink
```
Janiczek: `Shrink.noShrink expr`

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+
+        [] ->
+            Shrink.noShrink list
+
+
+shrinkTo : a -> Shrinker a
+shrinkTo shrunk _ =
+    True
+        |> Shrink.bool
+        |> Shrink.map (always shrunk)
```
Janiczek: I don't get what this function does. Is it something `Random.constant`-like?
If yes, does that mean that `If` shrinks like this?
```elm
If { test = _, then_ = a, else_ = b }
--> [ a, b ]
```
Does it shrink the `a` and `b` too? It seems to me like it doesn't but I might have missed something.

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+---
+
+We cannot write a type annotation here.
+The `LazyList a` type used by shrinkers is not exposed outside `elm-explorations/test`.
+
+    shrink : a -> Shrinker a -> Shrinker a
+
+-}
+shrink expr shrinker =
+    shrinker expr
```
Janiczek: As said above, let's remove this

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,329 @@
...
+-}
+lazyMap2 f la lb =
+    Shrink.map f la
+        |> Shrink.andMap lb
+
+
+concatShrink : List (Shrinker a) -> Shrinker a
+concatShrink shrinkers =
+    shrinkers
+        |> List.foldl Shrink.merge Shrink.noShrink
```
Janiczek: With these two functions, we can probably start our own `Shrink.Extra` module. Then they can be named `map2` and `concat`. (Also, the `shrinkTo`, which I'd name `constant` if it does what I think it does, can move there too!)

tests/InferTypesFuzz.elm
szabba: That is exactly the reason.

Janiczek: Let's write a comment here: something along the lines of "We don't want to shrink non-empty list to an empty one - that would change the type, eg. from `List Int` to `List a`"

tests/InferTypesTest.elm
```diff
@@ -1,106 +1,151 @@
...
-                        , located (Canonical.Literal (Literal.String "two"))
-                        , located (Canonical.Literal (Literal.Int 3))
-                        ]
-                  , Err (TypeMismatch Bool String)
-                  )
-                ]
+        [ runSection "list"
+            [ ( "empty list"
+              , Canonical.List []
+              , Ok (Type.List (Type.Var 1))
```
Janiczek: We can try to drop the `Type.` qualifiers in the expected outputs.
`Ok (List (Var 1))` etc.

tests/InferTypesFuzz.elm
szabba: The empty list is inferred to what dump as `List a`. We could generate `Canonical.List []` for `List (Var n)`, but I'm not certain it will always work out well with recursive generation. It might be ok but I have not thought it through.

Janiczek: Ah. Right, let's keep it as it is.

tests/InferTypesTest.elm
szabba: We could even go with `Ok "List a"` if we call `dumpType` in the test. :thinking:

Janiczek: I'd probably rather keep the `Type` type there, not call `dumpType` :) but yeah, we totally could!

tests/InferTypesFuzz.elm
szabba: Now that you've pointed it out I think the other cases might be wrong. AFAIR shrinkers should only do one step of shrinking at a time. elm-test is responsible for shrinking the output again as long it's possible. :man_facepalming:

tests/InferTypesFuzz.elm
szabba: I didn't think very hard about the module name. It's the fuzzing bits for `InferTypesFuzz`. Would the name make more sense if the fuzzed tests was moved inside this module? Or do we want to rename the module and keep it's scope?

Janiczek: Yeah moving the fuzz tests here would work!

tests/InferTypesTest.elm
szabba: The newest version here uses `dumpType`, just to see how that looks. Pro: it reads nice. Con:  breaking `dumpType` will break these tests too.

tests/InferTypesFuzz.elm
```diff
@@ -0,0 +1,391 @@
...
+                        |> Expect.equal (Ok typeWanted)
+
+        fuzzExpressions : String -> List Type -> Test
+        fuzzExpressions description types =
+            types
+                |> List.map fuzzExpr
+                |> describe description
+    in
+    describe "Stage.InferType"
+        [ describe "fuzz exprInfer"
```
Janiczek: `"fuzz inferExpr"`?

tests/InferTypesTest.elm
```diff
@@ -1,106 +1,120 @@
...
-                        [ located (Canonical.Literal (Literal.Bool True))
-                        , located (Canonical.Literal (Literal.String "two"))
-                        , located (Canonical.Literal (Literal.Int 3))
-                        ]
-                  , Err (TypeMismatch Bool String)
-                  )
-                ]
+        [ runSection "list"
+            [ ( "empty list"
+              , Unwrapped.List []
```
Janiczek: `Unwrapped` is a bit non-ideal - "which Unwrapped?"
If no qualifier or `Canonical` is not possible because of collisions, I'd go for `CanonicalU` or something like that.

Janiczek:
Looking good overall! :bowing_man: I'll wait for your signal - don't know if you want to do something more with the shrinkers or not.

tests/InferTypesFuzz.elm
szabba: Hope the new labelling is clearer!

tests/InferTypesTest.elm
Janiczek: I don't know... I see it looks nicer, but I'd still rather have the AST there as the expected output than the String.

PR #51, opened 2019-07-16
AST refactoring to extract position (Located Expr)

Warry:
The general assignment is adapting the AST (each stage) to support the following:
```elm
type Located expr
    = Located Region expr

type alias Region =
    { start : Position
    , end : Position
    }

type alias Position =
    { row : Int
    , col : Int
    }

located : Parser_ p -> Parser_ (Located p)
located p =
    P.succeed
        (\( row1, col1 ) expr_ ( row2, col2 ) ->
            Located.located
                { start = { row = row1, col = col1 }
                , end = { row = row2, col = col2 }
                }
                expr_
        )
        |= P.getPosition
        |= p
        |= P.getPosition
```

The objective, is to track position to display them into error messages.

- [x] Add Located Api
- [x] Support Located in Parser
- [x] Refactor Stages(..)
- [x] Refactor Tests
    - test now look a bit cryptic, thanks to the Located leaking everywhere
- [ ] Actually show the position (and maybe even the lines around in the code?)

### DIFF
--- a/src/compiler/AST/Canonical.elm
+++ b/src/compiler/AST/Canonical.elm
@@ -2,7 +2,6 @@ module AST.Canonical exposing
     ( Expr(..)
     , LocatedExpr
     , ProjectFields
-    , lambda
     , var
     )
 
@@ -53,16 +52,3 @@ var qualifier name =
         { qualifier = qualifier
         , name = name
         }
-
-
-lambda : VarName -> LocatedExpr -> LocatedExpr
-lambda argument body =
-    -- we use body's location
-    Located.map
-        (\_ ->
-            Lambda
-                { argument = argument
-                , body = body
-                }
-        )
-        body
--- a/src/compiler/AST/Canonical.elm
+++ b/src/compiler/AST/Canonical.elm
@@ -1,11 +1,13 @@
 module AST.Canonical exposing
     ( Expr(..)
+    , LocatedExpr
     , ProjectFields
     , lambda
     , var
     )
 
 import AST.Common.Literal exposing (Literal)
+import AST.Common.Located as Located exposing (Located)
 import Common.Types
     exposing
         ( Binding
@@ -17,7 +19,11 @@ import Dict.Any exposing (AnyDict)
 
 
 type alias ProjectFields =
-    { modules : Modules Expr }
+    { modules : Modules LocatedExpr }
+
+
+type alias LocatedExpr =
+    Located Expr
 
 
 {-| Differs from Frontend.Expr by:
@@ -30,15 +36,15 @@ type Expr
     = Literal Literal
     | Var { qualifier : ModuleName, name : VarName }
     | Argument VarName
-    | Plus Expr Expr
-    | Lambda { argument : VarName, body : Expr }
-    | Call { fn : Expr, argument : Expr }
-    | If { test : Expr, then_ : Expr, else_ : Expr }
-    | Let { bindings : AnyDict String VarName (Binding Expr), body : Expr }
-    | List (List Expr)
+    | Plus LocatedExpr LocatedExpr
+    | Lambda { argument : VarName, body : LocatedExpr }
+    | Call { fn : LocatedExpr, argument : LocatedExpr }
+    | If { test : LocatedExpr, then_ : LocatedExpr, else_ : LocatedExpr }
+    | Let { bindings : AnyDict String VarName (Binding LocatedExpr), body : LocatedExpr }
+    | List (List LocatedExpr)
     | Unit
-    | Tuple Expr Expr
-    | Tuple3 Expr Expr Expr
+    | Tuple LocatedExpr LocatedExpr
+    | Tuple3 LocatedExpr LocatedExpr LocatedExpr
 
 
 var : ModuleName -> VarName -> Expr
@@ -49,9 +55,14 @@ var qualifier name =
         }
 
 
-lambda : VarName -> Expr -> Expr
+lambda : VarName -> LocatedExpr -> LocatedExpr
 lambda argument body =
-    Lambda
-        { argument = argument
-        , body = body
-        }
+    -- we use body's location
+    Located.map
+        (\_ ->
+            Lambda
+                { argument = argument
+                , body = body
+                }
+        )
+        body
--- a/src/compiler/AST/Frontend.elm
+++ b/src/compiler/AST/Frontend.elm
@@ -59,23 +59,6 @@ lambda arguments body =
         }
 
 
-
-{- Let's not get ahead of ourselves
-
-   | Let
-       { varName : VarName
-       , varBody : LocatedExpr
-       , body : LocatedExpr
-       }
-   | Fixpoint LocatedExpr
-   | Operator
-       { opName : VarName
-       , left : LocatedExpr
-       , right : LocatedExpr
-       }
--}
-
-
 {-| A helper for the Transform library.
 -}
 recurse : (Expr -> Expr) -> Expr -> Expr
--- a/src/compiler/AST/Frontend.elm
+++ b/src/compiler/AST/Frontend.elm
@@ -1,5 +1,6 @@
 module AST.Frontend exposing
     ( Expr(..)
+    , LocatedExpr
     , ProjectFields
     , lambda
     , transform
@@ -7,6 +8,7 @@ module AST.Frontend exposing
     )
 
 import AST.Common.Literal exposing (Literal)
+import AST.Common.Located as Located exposing (Located)
 import Common
 import Common.Types
     exposing
@@ -19,22 +21,26 @@ import Transform
 
 
 type alias ProjectFields =
-    { modules : Modules Expr }
+    { modules : Modules LocatedExpr }
+
+
+type alias LocatedExpr =
+    Located Expr
 
 
 type Expr
     = Literal Literal
     | Var { qualifier : Maybe ModuleName, name : VarName }
     | Argument VarName
-    | Plus Expr Expr
-    | Lambda { arguments : List VarName, body : Expr }
-    | Call { fn : Expr, argument : Expr }
-    | If { test : Expr, then_ : Expr, else_ : Expr }
-    | Let { bindings : List (Binding Expr), body : Expr }
-    | List (List Expr)
+    | Plus LocatedExpr LocatedExpr
+    | Lambda { arguments : List VarName, body : LocatedExpr }
+    | Call { fn : LocatedExpr, argument : LocatedExpr }
+    | If { test : LocatedExpr, then_ : LocatedExpr, else_ : LocatedExpr }
+    | Let { bindings : List (Binding LocatedExpr), body : LocatedExpr }
+    | List (List LocatedExpr)
     | Unit
-    | Tuple Expr Expr
-    | Tuple3 Expr Expr Expr
+    | Tuple LocatedExpr LocatedExpr
+    | Tuple3 LocatedExpr LocatedExpr LocatedExpr
 
 
 var : Maybe ModuleName -> VarName -> Expr
@@ -45,7 +51,7 @@ var qualifier name =
         }
 
 
-lambda : List VarName -> Expr -> Expr
+lambda : List VarName -> LocatedExpr -> Expr
 lambda arguments body =
     Lambda
         { arguments = arguments
@@ -58,14 +64,14 @@ lambda arguments body =
 
    | Let
        { varName : VarName
-       , varBody : Expr
-       , body : Expr
+       , varBody : LocatedExpr
+       , body : LocatedExpr
        }
-   | Fixpoint Expr
+   | Fixpoint LocatedExpr
    | Operator
        { opName : VarName
-       , left : Expr
-       , right : Expr
+       , left : LocatedExpr
+       , right : LocatedExpr
        }
 -}
 
@@ -85,41 +91,43 @@ recurse f expr =
             expr
 
         Plus e1 e2 ->
-            Plus (f e1) (f e2)
+            Plus
+                (Located.map f e1)
+                (Located.map f e2)
 
         Lambda ({ body } as lambda_) ->
-            Lambda { lambda_ | body = f body }
+            Lambda { lambda_ | body = Located.map f body }
 
         Call { fn, argument } ->
             Call
-                { fn = f fn
-                , argument = f argument
+                { fn = Located.map f fn
+                , argument = Located.map f argument
                 }
 
         If { test, then_, else_ } ->
             If
-                { test = f test
-                , then_ = f then_
-                , else_ = f else_
+                { test = Located.map f test
+                , then_ = Located.map f then_
+                , else_ = Located.map f else_
                 }
 
         Let { bindings, body } ->
             Let
-                { bindings = List.map (Common.mapBinding f) bindings
-                , body = f body
+                { bindings = List.map (Common.mapBinding (Located.map f)) bindings
+                , body = Located.map f body
                 }
 
         List items ->
-            List (List.map f items)
+            List (List.map (Located.map f) items)
 
         Unit ->
             expr
 
         Tuple e1 e2 ->
-            Tuple (f e1) (f e2)
+            Tuple (Located.map f e1) (Located.map f e2)
 
         Tuple3 e1 e2 e3 ->
-            Tuple3 (f e1) (f e2) (f e3)
+            Tuple3 (Located.map f e1) (Located.map f e2) (Located.map f e3)
 
 
 transform : (Expr -> Expr) -> Expr -> Expr
--- a/src/compiler/AST/Frontend.elm
+++ b/src/compiler/AST/Frontend.elm
@@ -63,6 +63,10 @@ lambda arguments body =
 -}
 recurse : (Expr -> Expr) -> Expr -> Expr
 recurse f expr =
+    let
+        f_ =
+            Located.map f
+    in
     case expr of
         Literal _ ->
             expr
@@ -75,42 +79,42 @@ recurse f expr =
 
         Plus e1 e2 ->
             Plus
-                (Located.map f e1)
-                (Located.map f e2)
+                (f_ e1)
+                (f_ e2)
 
         Lambda ({ body } as lambda_) ->
-            Lambda { lambda_ | body = Located.map f body }
+            Lambda { lambda_ | body = f_ body }
 
         Call { fn, argument } ->
             Call
-                { fn = Located.map f fn
-                , argument = Located.map f argument
+                { fn = f_ fn
+                , argument = f_ argument
                 }
 
         If { test, then_, else_ } ->
             If
-                { test = Located.map f test
-                , then_ = Located.map f then_
-                , else_ = Located.map f else_
+                { test = f_ test
+                , then_ = f_ then_
+                , else_ = f_ else_
                 }
 
         Let { bindings, body } ->
             Let
-                { bindings = List.map (Common.mapBinding (Located.map f)) bindings
-                , body = Located.map f body
+                { bindings = List.map (Common.mapBinding f_) bindings
+                , body = f_ body
                 }
 
         List items ->
-            List (List.map (Located.map f) items)
+            List (List.map f_ items)
 
         Unit ->
             expr
 
         Tuple e1 e2 ->
-            Tuple (Located.map f e1) (Located.map f e2)
+            Tuple (f_ e1) (f_ e2)
 
         Tuple3 e1 e2 e3 ->
-            Tuple3 (Located.map f e1) (Located.map f e2) (Located.map f e3)
+            Tuple3 (f_ e1) (f_ e2) (f_ e3)
 
 
 transform : (Expr -> Expr) -> Expr -> Expr
--- a/src/compiler/AST/Typed.elm
+++ b/src/compiler/AST/Typed.elm
@@ -32,7 +32,7 @@ type alias ProjectFields =
     { modules : Modules LocatedExpr }
 
 
-{-| Differs from Canonical.LocatedExpr by:
+{-| Differs from Canonical.Expr by:
 
   - being a tuple of the underlying Expr\_ type and its type
 
--- a/src/compiler/AST/Typed.elm
+++ b/src/compiler/AST/Typed.elm
@@ -32,6 +32,10 @@ type alias ProjectFields =
     { modules : Modules LocatedExpr }
 
 
+type alias LocatedExpr =
+    Located Expr
+
+
 {-| Differs from Canonical.Expr by:
 
   - being a tuple of the underlying Expr\_ type and its type
@@ -39,10 +43,6 @@ type alias ProjectFields =
 TODO make this opaque, add accessors etc.
 
 -}
-type alias LocatedExpr =
-    Located Expr
-
-
 type alias Expr =
     ( Expr_, Type )
 
--- a/src/compiler/AST/Typed.elm
+++ b/src/compiler/AST/Typed.elm
@@ -8,7 +8,6 @@ module AST.Typed exposing
     , isArgument
     , lambda
     , let_
-    , mapExpr
     , recursiveChildren
     , transformAll
     , transformOnce
@@ -35,7 +34,7 @@ type alias ProjectFields =
 
 {-| Differs from Canonical.LocatedExpr by:
 
-  - being a tuple of the underlying LocatedExpr\_ type and its type
+  - being a tuple of the underlying Expr\_ type and its type
 
 TODO make this opaque, add accessors etc.
 
@@ -82,25 +81,21 @@ let_ bindings body =
         }
 
 
-
---TODO: Refactor the rest using Transform
-
-
 {-| A helper for the Transform library.
 -}
 recurse : (LocatedExpr -> LocatedExpr) -> LocatedExpr -> LocatedExpr
-recurse f expr =
+recurse f located =
     mapExpr
-        (\expr_ ->
-            case expr_ of
+        (\expr ->
+            case expr of
                 Literal _ ->
-                    expr_
+                    expr
 
                 Var _ ->
-                    expr_
+                    expr
 
                 Argument _ ->
-                    expr_
+                    expr
 
                 Plus e1 e2 ->
                     Plus
@@ -139,30 +134,30 @@ recurse f expr =
                     Tuple3 (f e1) (f e2) (f e3)
 
                 Unit ->
-                    expr_
+                    expr
         )
-        expr
+        located
 
 
 transformOnce : (LocatedExpr -> LocatedExpr) -> LocatedExpr -> LocatedExpr
-transformOnce pass expr_ =
+transformOnce pass located =
     Transform.transformOnce
         recurse
         pass
-        expr_
+        located
 
 
 transformAll : List (LocatedExpr -> Maybe LocatedExpr) -> LocatedExpr -> LocatedExpr
-transformAll passes expr_ =
+transformAll passes located =
     Transform.transformAll
         recurse
         (Transform.orList passes)
-        expr_
+        located
 
 
 isArgument : VarName -> LocatedExpr -> Bool
-isArgument name expr =
-    case getExpr expr of
+isArgument name located =
+    case getExpr located of
         Argument argName ->
             argName == name
 
@@ -171,8 +166,8 @@ isArgument name expr =
 
 
 recursiveChildren : (LocatedExpr -> List LocatedExpr) -> LocatedExpr -> List LocatedExpr
-recursiveChildren fn expr =
-    case getExpr expr of
+recursiveChildren fn located =
+    case getExpr located of
         Literal _ ->
             []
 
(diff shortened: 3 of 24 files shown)


src/compiler/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -39,8 +40,12 @@ import Stage.InferTypes.TypeEquation exposing (TypeEquation, equals)
 import Transform
 
 
-generateEquations : IdSource -> Typed.Expr -> ( List TypeEquation, IdSource )
-generateEquations idSource ( expr, type_ ) =
+generateEquations : IdSource -> Typed.LocatedExpr -> ( List TypeEquation, IdSource )
+generateEquations idSource located_ =
+    let
+        ( expr, type_ ) =
+            Located.unwrap located_
```
Janiczek: no need for `located_` I believe as there's nothing that would shadow the name `located`?

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -341,13 +350,15 @@ expr =
...
+                (Located.merge
+                    (\fn argument ->
+                        Frontend.Call
+                            { fn = fn
+                            , argument = argument
+                            }
+                    )
                 )
-            , PP.infixLeft 1 (P.symbol (P.Token "+" ExpectingPlusOperator)) Plus
+            , PP.infixLeft 1 (P.symbol (P.Token "+" ExpectingPlusOperator)) (Located.merge Plus)
```
Warry: Here it's merged, but nowhere else... either it shouldn't merge at all, or merge also in function, let, lists..

src/compiler/AST/Canonical.elm
```diff
@@ -49,9 +55,14 @@ var qualifier name =
...
 
 
-lambda : VarName -> Expr -> Expr
+lambda : VarName -> LocatedExpr -> LocatedExpr
 lambda argument body =
-    Lambda
-        { argument = argument
-        , body = body
-        }
+    -- we use body's location
```
Warry: Here we arbitrarily choose to keep the location of the body when optimizing. It should live in Optimization or be fixed.

src/compiler/Stage/InferTypes/AssignIds.elm
```diff
@@ -40,39 +40,52 @@ Output:
...
 import AST.Common.Type as Type
 import AST.Typed as Typed
 import Common
 import Dict.Any
 import Stage.InferTypes.IdSource as IdSource exposing (IdSource)
 
 
-assignIds : Canonical.Expr -> ( Typed.Expr, IdSource )
+assignIds : Canonical.LocatedExpr -> ( Typed.LocatedExpr, IdSource )
 assignIds expr =
```
Warry: use located like everywhere.

src/compiler/Stage/InferTypes/AssignIds.elm
```diff
@@ -40,39 +40,52 @@ Output:
...
     assignIdsWith IdSource.empty expr
 
 
-assignIdsWith : IdSource -> Canonical.Expr -> ( Typed.Expr, IdSource )
+assignIdsWith : IdSource -> Canonical.LocatedExpr -> ( Typed.LocatedExpr, IdSource )
 assignIdsWith idSource expr =
```
Warry: use located like everywhere.

src/compiler/Stage/Optimize.elm
```diff
@@ -12,34 +13,43 @@ optimize project =
...
 
         _ ->
             Nothing
 
 
-optimizeIfLiteralBool : Typed.Expr -> Maybe Typed.Expr
-optimizeIfLiteralBool ( expr, _ ) =
-    case expr of
+optimizeIfLiteralBool : Typed.LocatedExpr -> Maybe Typed.LocatedExpr
+optimizeIfLiteralBool expr =
```
Warry: use located like everywhere.

src/compiler/Stage/Optimize.elm
```diff
@@ -12,34 +13,43 @@ optimize project =
...
-optimizePlus : Typed.Expr -> Maybe Typed.Expr
-optimizePlus ( expr, _ ) =
-    case expr of
-        Typed.Plus ( Typed.Literal (Literal.Int left), _ ) ( Typed.Literal (Literal.Int right), _ ) ->
-            Just
-                ( Typed.Literal (Literal.Int (left + right))
-                , Type.Int
-                )
+optimizePlus : Typed.LocatedExpr -> Maybe Typed.LocatedExpr
+optimizePlus expr =
```
Warry: use located like everywhere else.

src/compiler/Stage/Parse/Parser.elm
```diff
@@ -324,7 +333,7 @@ topLevelDeclaration =
         |= expr
 
 
-expr : Parser_ Frontend.Expr
+expr : Parser_ Frontend.LocatedExpr
 expr =
```
Warry: use located like everywhere.

src/compiler/AST/Typed.elm
```diff
@@ -43,119 +52,127 @@ type Expr_
...
-            Tuple (f e1) (f e2)
 
-        Tuple3 e1 e2 e3 ->
-            Tuple3 (f e1) (f e2) (f e3)
-    , type_
-    )
+{-| A helper for the Transform library.
+-}
+recurse : (LocatedExpr -> LocatedExpr) -> LocatedExpr -> LocatedExpr
+recurse f expr =
```
Warry: use located like everywhere.

src/compiler/AST/Typed.elm
```diff
@@ -43,119 +52,127 @@ type Expr_
...
+                Unit ->
+                    expr_
+        )
+        expr
 
 
-transformOnce : (Expr -> Expr) -> Expr -> Expr
-transformOnce pass expr =
+transformOnce : (LocatedExpr -> LocatedExpr) -> LocatedExpr -> LocatedExpr
+transformOnce pass expr_ =
```
Warry: use located like everywhere .

src/compiler/AST/Typed.elm
```diff
@@ -43,119 +52,127 @@ type Expr_
...
         (Transform.orList passes)
-        expr
+        expr_
 
 
-isArgument : VarName -> Expr -> Bool
-isArgument name ( expr_, _ ) =
-    case expr_ of
+isArgument : VarName -> LocatedExpr -> Bool
+isArgument name expr =
```
Warry: use located like everywhere.

src/compiler/Stage/InferTypes.elm
```diff
@@ -33,7 +34,7 @@ inferTypes project =
         |> Result.mapError TypeError
 
 
-inferExpr : Canonical.Expr -> Result TypeError Typed.Expr
+inferExpr : Canonical.LocatedExpr -> Result TypeError Typed.LocatedExpr
 inferExpr expr =
```
Warry: use located like everywhere.

src/compiler/Stage/InferTypes/AssignIds.elm
```diff
@@ -40,39 +40,52 @@ Output:
...
+    )
+
+
+assignId : IdSource -> Typed.Expr_ -> ( Typed.Expr, IdSource )
+assignId idSource expr =
+    IdSource.one idSource (\id -> ( expr, Type.Var id ))
+
+
+assignIdsWithHelp : IdSource -> Canonical.Expr -> ( Typed.Expr, IdSource )
+assignIdsWithHelp idSource expr =
```
Warry: use located like everywhere.

src/compiler/Stage/Parse/Parser.elm
Janiczek: I don't know if that's true.
I think it should merge in all the `andThenOneOf` cases, but not in the `oneOf` cases :thinking: as those don't take a second Expr as an argument, right?

src/compiler/AST/Canonical.elm
Janiczek: Body's location is not enough though, right?
```
\x -> x + 1
```
Ideally we'd like it to start with the `\` and end with the `1` but if we use body's location, the location will only contain `x + 1` if I understand this correctly.

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -4,7 +4,7 @@ module Stage.Emit.JavaScript exposing (emitExpr, emitTopLevelDeclaration)
 
 import AST.Backend as Backend
 import AST.Common.Literal exposing (Literal(..))
-import AST.Typed exposing (Expr_(..))
+import AST.Typed exposing (Expr_(..), getExpr)
```
Janiczek: I'd rather not import `getExpr` like this but use it qualified (`AST.Typed.getExpr`)

src/compiler/Stage/InferTypes/AssignIds.elm
Janiczek: This comment seems incorrect? :sweat_smile:

src/compiler/AST/Frontend.elm
```diff
@@ -58,14 +64,14 @@ lambda arguments body =
...
+       , body : LocatedExpr
        }
-   | Fixpoint Expr
+   | Fixpoint LocatedExpr
    | Operator
        { opName : VarName
-       , left : Expr
-       , right : Expr
+       , left : LocatedExpr
+       , right : LocatedExpr
```
Janiczek: I know this is unrelated to your PR but can we please remove this comment?

src/compiler/AST/Typed.elm
```diff
@@ -25,16 +30,20 @@ import Transform
...
 type alias ProjectFields =
-    { modules : Modules Expr }
+    { modules : Modules LocatedExpr }
 
 
-{-| Differs from Canonical.Expr by:
+{-| Differs from Canonical.LocatedExpr by:
 
-  - being a tuple of the underlying Expr\_ type and its type
+  - being a tuple of the underlying LocatedExpr\_ type and its type
```
Janiczek: This seems like a search-and-replace gone too far. I'd rather move this comment under `Expr_` and make it talk about Expr_ instead of LocatedExpr.

src/compiler/AST/Typed.elm
```diff
@@ -43,119 +52,127 @@ type Expr_
...
-                { bindings = Dict.Any.map (always (Common.mapBinding f)) bindings
-                , body = f body
-                }
-
-        List items ->
-            List (List.map f items)
 
-        Unit ->
-            expr
+--TODO: Refactor the rest using Transform
```
Janiczek: Is this TODO current?

src/compiler/Error.elm
```diff
@@ -12,6 +12,7 @@ module Error exposing
     , toString
     )
 
+import AST.Common.Located exposing (Region)
```
Janiczek: Is this needed after all?

src/compiler/AST/Typed.elm
```diff
@@ -1,16 +1,21 @@
...
     ( Expr
     , Expr_(..)
+    , LocatedExpr
     , ProjectFields
+    , getExpr
+    , getType
     , isArgument
     , lambda
     , let_
+    , mapExpr
```
Janiczek: We currently don't need to expose this, `AST.Typed` is the only module using this

src/compiler/AST/Typed.elm
```diff
@@ -196,3 +213,18 @@ recursiveChildren fn ( expr, _ ) =
 
         Tuple3 e1 e2 e3 ->
             fn e1 ++ fn e2 ++ fn e3
+
+
+mapExpr : (a -> b) -> Located ( a, c ) -> Located ( b, c )
```
Janiczek: Let's make the types specific to document it better.
`mapExpr : (Expr_ -> Expr_) -> Located Expr -> Located Expr`

src/compiler/AST/Typed.elm
```diff
@@ -196,3 +213,18 @@ recursiveChildren fn ( expr, _ ) =
...
+mapExpr =
+    Located.map << Tuple.mapFirst
+
+
+getExpr : Located ( a, b ) -> a
```
Janiczek: similar here: `getExpr : LocatedExpr -> Expr_`
I'm a bit torn about whether to name it `getExpr` or `getExpr_` as it doesn't return `Expr` but `Expr_`

src/compiler/AST/Typed.elm
```diff
@@ -196,3 +213,18 @@ recursiveChildren fn ( expr, _ ) =
...
+getExpr =
+    Tuple.first << Located.unwrap
+
+
+getType : Located ( a, b ) -> b
```
Janiczek: and similar here: `getType : LocatedExpr -> Type`

src/compiler/AST/Typed.elm
Warry: Nice catch !

src/compiler/AST/Typed.elm
Warry: Nope, I'll remove it

src/compiler/Error.elm
Warry: It depends: do you want the displaying of the error on another PR ?

Janiczek: It can go in a different PR, this is big enough (and might get bigger depending on what we decide to do about the tests and Unwrapped variants)

src/compiler/AST/Frontend.elm
Warry: upcoming.

src/compiler/Stage/Parse/Parser.elm
Warry: I remember why I made it now. I used generated expressions so the tests told me this was right. I was really tired when commented last night ^^.

Janiczek: OK, then this code probably doesn't need changing :) marking as resolved.

src/compiler/Stage/Desugar.elm
```diff
@@ -195,9 +156,21 @@ desugarTuple3 recurse e1 e2 e3 =
     Canonical.Lambda arg1 (Canonical.Lambda arg2 body)
 
 -}
-curryLambda : List VarName -> Canonical.Expr -> Canonical.Expr
+curryLambda : List VarName -> Canonical.LocatedExpr -> Canonical.LocatedExpr
```
Janiczek: I believe this still has the problem of the child lambdas having the Location of the body, not the parent lambda.
```elm
\x y z -> x + y + z
-->
\x -> \y -> \z -> x + y + z
```
We want all three lambdas to have the same position: the one of the original big lambda, not of the body.

I believe this can be done by passing the original position as an argument to this `curryLambda`. The big `desugarExpr` function that calls this has access to that original position.

src/compiler/AST/Canonical.elm
Warry: For this, I transfered it to Desugar, its only reference.

src/compiler/AST/Canonical.elm
```diff
@@ -47,11 +52,3 @@ var qualifier name =
         { qualifier = qualifier
         , name = name
         }
-
-
-lambda : VarName -> Expr -> Expr
-lambda argument body =
```
Janiczek: Was there a good reason for removing this function? (Its only usage was Desugar.curryLambda I believe, but still)
There totally might have been and I just missed it, dunno...

src/compiler/AST/Frontend.elm
```diff
@@ -85,41 +74,43 @@ recurse f expr =
             expr
 
         Plus e1 e2 ->
-            Plus (f e1) (f e2)
+            Plus
+                (Located.map f e1)
```
Janiczek: We can probably do
```elm
recurse f expr =
  let
    f_ =
      Located.map f
  in
  ...
```
and then use that `f_`

src/compiler/AST/Typed.elm
```diff
@@ -25,16 +29,20 @@ import Transform
 
 
 type alias ProjectFields =
-    { modules : Modules Expr }
+    { modules : Modules LocatedExpr }
 
 
-{-| Differs from Canonical.Expr by:
+{-| Differs from Canonical.LocatedExpr by:
```
Janiczek: This comment still needs to move next to `Expr` and to talk about `Expr`

src/compiler/AST/Frontend.elm
Janiczek: Also wait a minute ... Frontend `recurse` works on `Expr` but Typed `recurse` works on `LocatedExpr` :thinking:  Which one is right?

src/compiler/Stage/Desugar.elm
```diff
@@ -262,11 +235,58 @@ qualifiedVarInImportedModule modules maybeModuleName varName =
...
             Maybe.andThen (Common.unalias thisModule) maybeModuleName
     in
     -- Reusing the existing functionality. TODO is this a good idea?
     qualifiedVarInImportedModule modules unaliasedModuleName varName
+
+
+locatedResultReturn : Frontend.LocatedExpr -> Canonical.Expr -> Result DesugarError Canonical.LocatedExpr
+locatedResultReturn located expr =
+    Ok
+        (Located.map
```
Janiczek: These wrappers call for some non-map helper like

```elm
Located.replaceWith : b -> Located a -> Located b
Located.replaceWith new (Located region _) =
  Located region new
```

Then this can be nicer:
```elm
Ok <| Located.replaceWith expr located
```
similarly for the other functions

src/compiler/AST/Typed.elm
```diff
@@ -32,7 +32,7 @@ type alias ProjectFields =
     { modules : Modules LocatedExpr }
 
 
-{-| Differs from Canonical.LocatedExpr by:
+{-| Differs from Canonical.Expr by:
```
Janiczek: still at the wrong place :upside_down_face: 

instead of
```elm
{-| ... -}
type alias LocatedExpr

type alias Expr
```
I wanted
```elm
type alias LocatedExpr

{-| ... -}
type alias Expr
```
Sorry, the language barrier is probably showing :) :czech_republic: :fr:

src/compiler/Stage/Desugar.elm
```diff
@@ -195,9 +156,21 @@ desugarTuple3 recurse e1 e2 e3 =
...
 -}
-curryLambda : List VarName -> Canonical.Expr -> Canonical.Expr
-curryLambda arguments body =
-    List.foldr Canonical.lambda body arguments
+curryLambda : Canonical.LocatedExpr -> List VarName -> Canonical.LocatedExpr -> Canonical.LocatedExpr
+curryLambda located arguments body =
+    List.foldr
+        (\argument body_ ->
+            Located.map
+                (\_ ->
```
Janiczek: this will also be able to use `Located.replaceWith`

src/compiler/AST/Frontend.elm
Warry: it should be Frontend, it's only one Located.map from Parser that should be done in Frontend. Typed can't, because it has to deal with modules which carries Located

src/compiler/Stage/InferTypes/AssignIds.elm
```diff
@@ -40,39 +40,52 @@ Output:
...
     let
-        wrap : IdSource -> Typed.Expr_ -> ( Typed.Expr, IdSource )
-        wrap idSource_ expr_ =
-            IdSource.one idSource_ (\id -> ( expr_, Type.Var id ))
+        ( located_, idSource_ ) =
+            assignIdsWithHelp idSource (Located.unwrap located)
     in
-    case expr of
+    {- Keep location, for error context -}
+    ( Located.map (\_ -> located_) located
```
Janiczek: The `located` name can be renamed to `locatedCanonicalExpr`
The `located_` name probably needs to change to `typedExpr` as it's not really located
And then here we can write `Located.replaceWith typedExpr locatedCanonicalExpr` instead of the map

src/compiler/AST/Frontend.elm
Janiczek: yeah I've checked more in depth and looks like the types are alright after all :+1: it just raised a red flag in my head when I saw it first :)

src/compiler/Stage/Desugar.elm
```diff
@@ -262,11 +234,50 @@ qualifiedVarInImportedModule modules maybeModuleName varName =
...
+
+
+locatedResultMap3 : Frontend.LocatedExpr -> (a -> b -> c -> Canonical.Expr) -> Result DesugarError a -> Result DesugarError b -> Result DesugarError c -> Result DesugarError Canonical.LocatedExpr
+locatedResultMap3 located fn =
+    Result.map3
+        (\expr1 expr2 expr3 ->
+            Located.replaceWith
+                (fn expr1 expr2 expr3)
+                located
+        )
```
Janiczek: I'm just flagging these functions to later experiment with whether it wouldn't be nicer to inline them in the `let ...` they're used in. That doesn't need to happen in this PR though - I can do that on my own later.

src/compiler/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -297,19 +305,19 @@ generateEquations idSource ( expr, type_ ) =
             )
 
 
-findArgumentUsages : VarName -> Typed.Expr -> List Typed.Expr
-findArgumentUsages argument bodyExpr =
-    bodyExpr
+findArgumentUsages : VarName -> Typed.LocatedExpr -> List Typed.LocatedExpr
+findArgumentUsages argument located =
+    located
```
Janiczek: I think the previous naming here was more informative. I'd keep `bodyExpr` there.

src/compiler/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -297,19 +305,19 @@ generateEquations idSource ( expr, type_ ) =
...
         |> Transform.children Typed.recursiveChildren
         |> List.filter (Typed.isArgument argument)
 
 
-generateArgumentUsageEquations : Int -> List Typed.Expr -> List TypeEquation
-generateArgumentUsageEquations argumentId usages =
+generateArgumentUsageEquations : Int -> List Typed.LocatedExpr -> List TypeEquation
+generateArgumentUsageEquations argumentId exprs =
```
Janiczek: Same here - `exprs` loses context compared to `usages` or `argumentUsages`

src/compiler/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -297,19 +305,19 @@ generateEquations idSource ( expr, type_ ) =
...
     let
         argumentType =
             Type.Var argumentId
     in
     List.map
-        (\( _, usageType ) -> equals usageType argumentType)
-        usages
+        (\l -> equals (Typed.getType l) argumentType)
```
Janiczek: I'd either do
```elm
(Typed.getType >> equals argumentType)
```
or
```elm
(\usage -> equals (Typed.getType usage) argumentType)`
```

src/compiler/Stage/Optimize.elm
```diff
@@ -12,34 +13,41 @@ optimize project =
     Boilerplate.optimizeProject optimizeExpr project
 
 
-optimizeExpr : Typed.Expr -> Typed.Expr
-optimizeExpr expr =
+optimizeExpr : Typed.LocatedExpr -> Typed.LocatedExpr
+optimizeExpr =
```
Janiczek: Please let's keep the argument in the function definition, like it was. I know it can be eta-reduced, but that goes contrary to the readability / simplicity / understandability values. I've even written a blogpost about stuff related to that a few years ago :smile: http://martin.janiczek.cz/clanek/being-clever-antipattern-in-elm/

src/compiler/Stage/Optimize.elm
```diff
@@ -12,34 +13,41 @@ optimize project =
...
+    case Typed.getExpr located of
+        Typed.Plus l r ->
+            case ( Typed.getExpr l, Typed.getExpr r ) of
+                ( Typed.Literal (Literal.Int left), Typed.Literal (Literal.Int right) ) ->
+                    Just
+                        (Located.replaceWith
+                            ( Typed.Literal (Literal.Int (left + right))
+                            , Type.Int
+                            )
+                            located
```
Janiczek: It's a bit sad that something previously so simple now is so bloated with the location info, but yeah I don't know how to make it better. :thinking:

tests/EmitTest.elm
```diff
@@ -16,6 +17,37 @@ import Stage.Emit.JavaScript as JS
...
+    Located.located
+        -- position do not matters in emit
+        { start = { row = 0, col = 0 }, end = { row = 0, col = 0 } }
+        expr
+
+
+typed : Typed.Expr_ -> Typed.LocatedExpr
+typed expr =
+    Located.located
+        -- position do not matters in emit
```
Janiczek: I think we can remove both of those comments, and extract that position to some top-level `dummyPosition = { ... }`

tests/EmitTest.elm
```diff
@@ -29,30 +61,12 @@ javascript =
...
-            typedString : String -> Typed.Expr
-            typedString str =
-                typed (Literal (String str))
-
-            typedBool : Bool -> Typed.Expr
-            typedBool bool =
-                typed (Literal (Bool bool))
           in
-          describe "emitExpr"
+          describe "emitExpr_"
```
Janiczek: looking at what gets run inside `runTest`, this should probably be `emitExpr` instead of `emitExpr_`

tests/EmitTest.elm
```diff
@@ -29,30 +61,12 @@ javascript =
...
             [ describe "Int"
                 (List.map runTest
                     [ ( "positive int", Literal (Int 42), "42" )
                     , ( "negative int", Literal (Int -998), "-998" )
-                    , -- Elm wat
-                      ( "negative zero int", Literal (Int (negate 0)), "0" )
```
Janiczek: I'd like this test to come back :grimacing: similarly with the comment in the float test

tests/OptimizeTest.elm
```diff
@@ -16,11 +17,29 @@ import Stage.Optimize
...
+typed expr =
+    Located.located
+        -- position do not matters in optimize
+        { start = { row = 0, col = 0 }, end = { row = 0, col = 0 } }
+        expr
+
+
+typedInt : Int -> Typed.LocatedExpr
+typedInt int =
+    typed ( Literal (Int int), Type.Int )
```
Janiczek: making a note here to later try extracting all these helpers into some separate module that can be used by all those test files that define it themselves now...

tests/InferTypesTest.elm
```diff
@@ -30,86 +31,46 @@ typeInference =
         (List.map runSection
             [ ( "list"
               , [ ( "empty list"
-                  , Canonical.List []
-                  , Ok ( Typed.List [], Type.List (Type.Var 1) )
+                  , located { end = { col = 3, row = 1 }, start = { col = 1, row = 1 } } (Canonical.List [])
+                  , Ok (located { end = { col = 3, row = 1 }, start = { col = 1, row = 1 } } ( Typed.List [], List (Var 1) ))
```
Janiczek: Can we try modifying `runTest` to `Typed.getType`, so that we wouldn't have to show all that location info in the tests?

We would then not really test the expression themselves (eg. that `Canonical.List [a,b,c]` changes into `Typed.List [a,b,c]`), but I think the main thing is to test the type inference, thus the types... And we'd get rid of the location info! :tada:

```elm
input
  |> Stage.InferTypes.inferExpr
  |> Typed.getType
  |> Expect.equal output
```

tests/ParserTest.elm
```diff
@@ -439,408 +440,347 @@ expr =
...
                   , "\\x -> x + 1"
-                  , Just
-                        (AST.Frontend.lambda
-                            [ VarName "x" ]
-                            (Plus
-                                (Argument (VarName "x"))
-                                (Literal (Int 1))
-                            )
-                        )
+                  , Just (located { end = { col = 12, row = 1 }, start = { col = 1, row = 1 } } (Lambda { arguments = [ VarName "x" ], body = located { end = { col = 12, row = 1 }, start = { col = 7, row = 1 } } (Plus (located { end = { col = 8, row = 1 }, start = { col = 7, row = 1 } } (Argument (VarName "x"))) (located { end = { col = 12, row = 1 }, start = { col = 11, row = 1 } } (Literal (Int 1)))) }))
```
Janiczek: I don't see a nice way to get rid of the position info here. :thinking: The closest that comes to mind is having the Frontend.Expr type take a parameter of what its children are ... and then mapping from `Frontend.Expr Frontend.LocatedExpr` into `Frontend.Expr Frontend.Expr` (which btw won't typecheck without using something like recursion schemes - https://blog.sumtypeofway.com/an-introduction-to-recursion-schemes/) ... I can try that in my own later. Let's sleep on it :man_shrugging:

tests/InferTypesTest.elm
Warry: Located is within the Expr(_) type, so unless we unwrap recursively I don't think we can get rid of those.

Janiczek: if we do `Typed.getType`, we get rid of all Exprs, so no LocatedExprs either anymore

tests/EmitTest.elm
```diff
@@ -62,8 +81,7 @@ javascript =
                     [ ( "positive float", Literal (Float 12.3), "12.3" )
                     , ( "negative float", Literal (Float -12.3), "-12.3" )
                     , ( "positive zero float", Literal (Float 0.0), "0" )
-                    , -- Elm wat
-                      ( "negative zero float", Literal (Float -0.0), "0" )
+                    , ( "negative zero float", Literal (Float -0.0), "0" )
```
Janiczek: Comment that was previously there is missing. I want the reader of the tests to notice that this is unexpected output, but -- that's how the official Elm compiler works currently.

src/compiler/AST/Canonical.elm
Janiczek: @Warry Thoughts?

Warry:
\o/ I made second contributors with 7 commits. happyness :)

Issue #29, opened 2019-07-04
Add `(::)` operator for lists


Janiczek:
Should be similar to how `Plus` is done. **Keep it a separate case in the `Frontend.Expr`** - later we can try to think of a way to abstract all binary operators into something dynamic, read from source instead of baked into the compiler.

- [ ] add parser tests
- [ ] add a parser
- [ ] add optimization tests
  - [ ] if the right item (the list) is a literal list (not a variable for example), put the left item into it:
  - `myStuff :: [1, 2]` should become `[myStuff, 1, 2]`
- [ ] add optimization
- [ ] change how Lists are emitted: instead of `[1,2,3]` do something like `List$fromJSArray([1,2,3])`?
- [ ] add emit tests - `(::)` should be just a call of Elm function with those two params
- [ ] add emit

PR #52, opened 2019-07-16
Add `(++)` operator for lists

### LINKED ISSUE #30: Add `(++)` operator for lists
Should be similar to how `Plus` is done. **Keep it a separate case in the `Frontend.Expr`** - later we can try to think of a way to abstract all binary operators into something dynamic, read from source instead of baked into the compiler.

- [ ] add parser tests
- [ ] add a parser
- [ ] add optimization tests
  - [ ] if both of the items are literal lists (not a variable for example), combine them into one list:
  - `[1, 2] ++ [3, 4]` should become `[1, 2, 3, 4]`
- [ ] add optimization
- [ ] add emit tests -- TODO rethink this similar to #29
  - [ ] `[1] ++ [2]` should become `[1].concat([2])`
  - [ ] `stuff ++ [1]` should become `stuff.concat([1])`
  - [ ] `[1] ++ stuff` should become `[1].concat(stuff)`
- [ ] add emit

## Not in scope :x: :x: :x:

- [ ] appendable typeclass
- [ ] `(++)` for strings

discoCommando:
fixes #30

### DIFF
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -45,9 +45,6 @@ emitExpr ( expr, _ ) =
         Plus e1 e2 ->
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
-        ListConcat e1 e2 ->
-            "(" ++ emitExpr e1 ++ " ++ " ++ emitExpr e2 ++ ")"
-
         Lambda { argument, body } ->
             -- TODO are these parentheses needed?
             "((" ++ mangleVarName argument ++ ") => " ++ emitExpr body ++ ")"
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -4,7 +4,7 @@ module Stage.Emit.JavaScript exposing (emitExpr, emitTopLevelDeclaration)
 
 import AST.Backend as Backend
 import AST.Common.Literal exposing (Literal(..))
-import AST.Typed exposing (Expr_(..))
+import AST.Typed as Typed exposing (Expr_(..))
 import Common.Types
     exposing
         ( ModuleName(..)
@@ -14,9 +14,9 @@ import Common.Types
 import Dict.Any
 
 
-emitExpr : Backend.Expr -> String
-emitExpr ( expr, _ ) =
-    case expr of
+emitExpr : Backend.LocatedExpr -> String
+emitExpr located =
+    case Typed.getExpr located of
         Literal (Int int) ->
             String.fromInt int
 
@@ -81,7 +81,7 @@ emitExpr ( expr, _ ) =
             "[" ++ emitExpr e1 ++ "," ++ emitExpr e2 ++ "," ++ emitExpr e3 ++ "]"
 
 
-emitTopLevelDeclaration : TopLevelDeclaration Backend.Expr -> String
+emitTopLevelDeclaration : TopLevelDeclaration Backend.LocatedExpr -> String
 emitTopLevelDeclaration { module_, name, body } =
     "const "
         ++ mangleQualifiedVar module_ name
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -45,6 +45,9 @@ emitExpr ( expr, _ ) =
         Plus e1 e2 ->
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
+        ListConcat e1 e2 ->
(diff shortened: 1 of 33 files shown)


src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -45,6 +45,9 @@ emitExpr ( expr, _ ) =
         Plus e1 e2 ->
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
+        ListConcat e1 e2 ->
+            "(" ++ emitExpr e1 ++ " ++ " ++ emitExpr e2 ++ ")"
```
Janiczek: We'll have to do something different here. This is not valid JS. Also we'll have to break out of the "Lists represented by JS arrays" mindset, as hinted at in #29.

I think this could potentially be just a `Call` to `List.concat` of these two `Expr`s. I'm not sure whether to add this as an optimization step (`ListConcat e1 e2 -> Call (Call concat e1) e2`) or just here in Emit (`ListConcat e1 e2 -> emitExpr (Call ...)`)

Let me sleep on it :sweat_smile:

discoCommando: oh ok, my bad, I misunderstood this part of the code :)

Janiczek: no worries :)

tests/ParserTest.elm
```diff
@@ -813,6 +813,14 @@ expr =
...
                   , Just (List [ Literal (Int 1), Literal (Int 2), Literal (Int 3) ])
                   )
+                , ( "list concat"
+                  , "[] ++ []"
+                  , Just (ListConcat (List []) (List []))
+                  )
+                , ( "list concat did not mess up the simple addition"
+                  , "1 + 2"
+                  , Just (Plus (Literal <| Int 1) (Literal <| Int 2))
+                  )
```
Janiczek: Normally I'd be against this kind of test, but we don't have any tests for `+` so let's keep it, good job :) Anyway in the future this will have to change into some kind of `BinOp "+" ... ...`.

tests/InferTypesTest.elm
```diff
@@ -57,6 +57,20 @@ typeInference =
                   , Canonical.List [ Canonical.List [ Canonical.Literal (Int 1) ], Canonical.List [ Canonical.Literal (Bool False) ] ]
                   , Err (Error.TypeMismatch Type.Int Type.Bool)
                   )
+                , ( "List Concat"
+                  , Canonical.ListConcat (Canonical.List []) (Canonical.List [ Canonical.Literal (Bool True) ])
+                  , Ok ( Typed.ListConcat ( Typed.List [], Type.List <| Type.Bool ) ( Typed.List [ ( Typed.Literal (Bool True), Type.Bool ) ], Type.List <| Type.Bool ), Type.List <| Type.Bool )
```
Janiczek: extraneous `<|` in a few places (ie. `Type.List <| Type.Bool` can be `Type.List Type.Bool`)

tests/InferTypesTest.elm
```diff
@@ -57,6 +57,20 @@ typeInference =
...
+                  )
+                , ( "List Concat for list of int and int"
+                  , Canonical.ListConcat (Canonical.List [ Canonical.Literal (Int 1) ]) (Canonical.Literal (Int 2))
+                  , Err (Error.TypeMismatch (Type.List <| Type.Var 4) Type.Int)
+                  )
+
+                -- Probably we want a better way of testing polymorphic types (instead of specifying an Int like Var 4)
```
Janiczek: This comment is meant for the test above?
The `Type.List <| Type.Var 4` worries me. There is enough information in the input expression for the type inference algorithm to find out it's `List Int`. Can you update this test to show that?

Then that failing test will probably guide us to fix the type inference for `ListConcat` :slightly_smiling_face:

src/compiler/Stage/Emit/JavaScript.elm
Janiczek: OK so I think the best way to go about this will be desugaring it into `Call`. That means `AST.Canonical.Expr` and `AST.Typed.Expr` won't have `Binop` constructor!

Janiczek:
@discoCommando Let's try having the `ListConcat` only in the `AST.Frontend.Expr`, and desugaring it into `Call` so that the other phases don't need to think about `ListConcat` at all. That should also make this PR a lot smaller as a lot of the code will not be needed after that! :)

discoCommando:
Nice idea with desugaring as a function. Perhaps in the future we could just have a map `operator` -> Variable name of function?

Janiczek:
@discoCommando I think that is exactly what happens in official Elm.
ie. see all these: https://github.com/elm/core/blob/1.0.2/src/Basics.elm#L71-L89

Janiczek:
the binary operator abstraction is there just to make sure the parentheses are correct I believe... and after you're done with parentheses, you can think of it as function calls

PR #53, opened 2019-07-25
Tuple parser

### LINKED ISSUE #35: Implement parser for tuples and 3-tuples
Implement parser for tuples and 3-tuples so that the tests created in #34 are green! :green_heart:

halfzebra:
Closes #35

### DIFF
--- a/src/compiler/Error.elm
+++ b/src/compiler/Error.elm
@@ -60,7 +60,7 @@ type ParseContext
     | InLambda
     | InList
     | InUnit
-    | InTuple2
+    | InTuple
     | InTuple3
 
 
--- a/src/compiler/Error.elm
+++ b/src/compiler/Error.elm
@@ -60,6 +60,8 @@ type ParseContext
     | InLambda
     | InList
     | InUnit
+    | InTuple2
+    | InTuple3
 
 
 type ParseProblem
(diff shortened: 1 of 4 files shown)


src/compiler/Error.elm
```diff
@@ -60,6 +60,8 @@ type ParseContext
     | InLambda
     | InList
     | InUnit
+    | InTuple2
```
Janiczek: Let's name the normal tuple `Tuple` and the triple `Tuple3` :) No need for the `2`.

halfzebra: I've been considering proposing using the notation the original compiler uses and updating all the other types, what do you think?

Janiczek: @halfzebra I might be a bit confused now - you mean like this? https://github.com/elm/compiler/blob/master/compiler/src/AST/Source.hs#L66

halfzebra: I was thinking about the notation used for [JS constructors](https://github.com/elm/compiler/blob/33f153eea8222b481fcf2f6d42c83b1230d00d06/compiler/src/Generate/JavaScript/Expression.hs#L171).

I'll make it consistent first 🙂

Janiczek:
Thanks, great work! We can probably refactor `backtrackable` out later, but to be fair I tried to implement this without it and failed miserably. So I'm glad for your working version :)

halfzebra:
Thanks for merging! :+1:

PR #54, opened 2019-07-27
Support for cons (::) operator

### LINKED ISSUE #29: Add `(::)` operator for lists

halfzebra:
Closes #29 

- [x] add parser tests
- [x] add a parser
- [x] add optimization tests
    if the right item (the list) is a literal list (not a variable for example), put the left item into it:
    myStuff :: [1, 2] should become [myStuff, 1, 2]
- [x] add optimization
- [x] ~~change how Lists are emitted: instead of [1,2,3] do something like List$fromJSArray([1,2,3])?~~
- [x] add emit tests - (::) should be just a call of Elm function with those two params
- [x] add emit
- [x] type inference

### DIFF
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -47,7 +47,7 @@ emitExpr located =
 
         Cons e1 e2 ->
             -- TODO implement emitter for (::)
-            "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")" 
+            "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")"
 
         Lambda { argument, body } ->
             -- TODO are these parentheses needed?
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -46,7 +46,7 @@ emitExpr located =
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
         Cons e1 e2 ->
-            "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")"
+            "[" ++ emitExpr e1 ++ "].concat(" ++ emitExpr e2 ++ ")"
 
         Lambda { argument, body } ->
             -- TODO are these parentheses needed?
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -45,6 +45,10 @@ emitExpr located =
         Plus e1 e2 ->
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
+        Cons e1 e2 ->
+            -- TODO implement emitter for (::)
+            ""
+
         Lambda { argument, body } ->
             -- TODO are these parentheses needed?
             "((" ++ mangleVarName argument ++ ") => " ++ emitExpr body ++ ")"
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -46,7 +46,6 @@ emitExpr located =
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
         Cons e1 e2 ->
-            -- TODO implement emitter for (::)
             "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")"
 
         Lambda { argument, body } ->
--- a/src/compiler/Stage/Emit/JavaScript.elm
+++ b/src/compiler/Stage/Emit/JavaScript.elm
@@ -47,7 +47,7 @@ emitExpr located =
 
         Cons e1 e2 ->
             -- TODO implement emitter for (::)
-            ""
+            "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")" 
 
         Lambda { argument, body } ->
             -- TODO are these parentheses needed?
--- a/src/compiler/Stage/Optimize.elm
+++ b/src/compiler/Stage/Optimize.elm
@@ -17,6 +17,7 @@ optimizeExpr : Typed.LocatedExpr -> Typed.LocatedExpr
 optimizeExpr located =
     Typed.transformAll
         [ optimizePlus
+        , optimizeCons
         , optimizeIfLiteralBool
         ]
         located
@@ -38,6 +39,26 @@ optimizePlus located =
                             )
                             located
                         )
+
+                _ ->
+                    Nothing
+
+        _ ->
+            Nothing
+
+
+log : String -> a -> b -> b
+log label data returned =
+    Debug.log label data |> (\_ -> returned)
+
+
+optimizeCons : Typed.LocatedExpr -> Maybe Typed.LocatedExpr
+optimizeCons located =
+    case Typed.getExpr located of
+        Typed.Cons l r ->
+            case Typed.getExpr r of
+                Typed.List list ->
+                    Just (Typed.mapExpr (\_ -> Typed.List (l :: list)) r)
 
                 _ ->
                     Nothing
--- a/src/compiler/Stage/Optimize.elm
+++ b/src/compiler/Stage/Optimize.elm
@@ -47,11 +47,6 @@ optimizePlus located =
             Nothing
 
 
-log : String -> a -> b -> b
-log label data returned =
-    Debug.log label data |> (\_ -> returned)
-
-
 optimizeCons : Typed.LocatedExpr -> Maybe Typed.LocatedExpr
 optimizeCons located =
     case Typed.getExpr located of
--- a/src/compiler/Stage/Optimize.elm
+++ b/src/compiler/Stage/Optimize.elm
@@ -32,13 +32,7 @@ optimizePlus located =
         Typed.Plus l r ->
             case ( Typed.getExpr l, Typed.getExpr r ) of
                 ( Typed.Literal (Literal.Int left), Typed.Literal (Literal.Int right) ) ->
-                    Just
-                        (Located.replaceWith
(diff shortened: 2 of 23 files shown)


Janiczek:
@halfzebra How far you want to go with this ticket? All the checkboxes?

halfzebra:
Thanks for the feedback!

@Janiczek  I'd like to implement the whole thing if that's okay.

halfzebra:
@Janiczek how should emitted JavaScript look?

> add emit tests - `(::)` should be just a call of Elm function with those two params

Do you want to follow the original implementation of Lists? [Elm/Kernel/List.js](https://github.com/elm/core/blob/master/src/Elm/Kernel/List.js)?

Janiczek:
I'm thinking about it. This ties in with how do we want to override official packages' kernel modules for other platforms: https://ptb.discordapp.com/channels/578305644780716039/578305645824966669/602188453530173475
and I worry we'll have to flesh that out first to conclusively solve this.

In the meantime, because we're emitting lists to be native JS arrays, can you please do
`1 :: [2,3]` :arrow_right: `[2,3].slice().unshift(1)`
? (The `.slice()` is probably not needed for this particular example, but for `1 :: a` it would be)

and disregard this one?
> change how Lists are emitted: instead of [1,2,3] do something like List$fromJSArray([1,2,3])?

I'm going to create a channel on the Discord where we can discuss the overrides (Elm-only where possible and target platform for where not) and how to do them.

halfzebra:
I've experimented a bit with emitting, it seems like chaining `.unshift` does not return the array, so I went with `.concat` to implement a simple immutable alternative.

Janiczek:
Ah, didn't think of `unshift()` not returning the new array. Thanks!

halfzebra:
I think I'm lacking type inference for `(::)` and that's pretty much it 🤔 

Is there any particular reason for #54 not having this in a checklist?

Janiczek:
@halfzebra Do you mean #29 ? No reason, I think I just forgot it. Go for it :slightly_smiling_face:

halfzebra:
Gotcha, I'll get to it then 👍

halfzebra:
I've added two simplest tests for `(+)` and `(::)` just to see how it works, but now I'm a bit confused on how to go about making more test scenarios.

Any hints?

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -47,7 +47,7 @@ emitExpr located =
 
         Cons e1 e2 ->
             -- TODO implement emitter for (::)
-            ""
+            "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")"
```
harrysarson: Not pretty but is nice :+1:

halfzebra:
@harrysarson Thanks for the review!

I'm not quite sure if the type inference works as it should, what do you think?

harrysarson:
Ooh okay, I do not see any reason why it should not work but I admit to being out my depth. :shrug:

Janiczek:
I'm gonna take a look :)

Janiczek:
Running a simple "end-to-end" test:
1. `vim example-project/src/Main.elm`
```elm
module Main exposing (main)

main =
    1 :: 2 :: [ 3, 4 ]
```
2. `make`
```
---------------------------
-- STARTING THE COMPILER --
---------------------------
Main.main: Located { end = { col = 23, row = 5 }, start = { col = 5, row = 5 } } (Cons (Located { end = { col = 6, row = 5 }, start = { col = 5, row = 5 } } (Literal (Int 1))) (Located { end = { col = 23, row = 5 }, start = { col = 10, row = 5 } } (Cons (Located { end = { col = 11, row = 5 }, start = { col = 10, row = 5 } } (Literal (Int 2))) (Located { end = { col = 23, row = 5 }, start = { col = 15, row = 5 } } (List [Located { end = { col = 18, row = 5 }, start = { col = 17, row = 5 } } (Literal (Int 3)),Located { end = { col = 21, row = 5 }, start = { col = 20, row = 5 } } (Literal (Int 4))])))))
Compilation finished, writing output to `out.js`.
---------------------------
-- WRITING TO FS ----------
---------------------------
const Main$main = [1, 2, 3, 4];
```
That looks fine! :+1:
I'll look at type inference, maybe write some tests, and then do a code review

Janiczek:
```elm
main =
    (1 :: 2) :: [ 3, 4 ]
```
should fail (`List a :: List a` instead of `a :: List a`, so to speak) but doesn't:
```
const Main$main = [[].concat([1],2), 3, 4];
```

Janiczek:
I've added the failing tests to the test suite

halfzebra:
Thanks for the tests, I'd like to give it a shot tomorrow!

Janiczek:
Note I'm actually not sure if the expected error I wrote in the last test is correct - I took a guess - but it definitely should not return `Ok` :sweat_smile:

halfzebra:
I think I've made some progress, but the test with type mismatch yields a pretty strange error:

```
✗ fail with wrong argument types

    Err (TypeMismatch (List (Var 7)) Int)
    ╷
    │ Expect.equal
    ╵
    Err (TypeMismatch (List Int) (List (List Int)))
``` 

Any hints on fixing this?

Janiczek:
Great that it's at least throwing a type mismatch error currently! I'm looking at the issue you mentioned :upside_down_face: 

debugging

Janiczek:
From `t0 == Int` in the equations and the actual error (List t0 /= Int) I believe it has enough information to go to the `List Int /= Int` phase (which is probably correct output instead of my guessed `List (List Int) /= List Int`).

Maybe we need it to unify whatever it can *after* the generation of the substitution map fails, too?

ie. it stopped before doing the `List t0 -> List Int` substitution, but we want it to finish that if it's a part of the error.

Janiczek:
Fixed that for you :)

Janiczek:
@halfzebra From your point, is there anything left to do before a code review?
Also, do you think this comment is OK? https://github.com/elm-in-elm/compiler/pull/54/commits/4a54b96c76e578b128fa6787de00d67f38c6c310#diff-a8128af301102f94c83ad003e4c04350R87-R105

(It's that "finish unifying types even in the error case" thing from a few comments earlier)

src/compiler/Stage/Optimize.elm
```diff
@@ -46,6 +47,26 @@ optimizePlus located =
             Nothing
 
 
+log : String -> a -> b -> b
+log label data returned =
+    Debug.log label data |> (\_ -> returned)
```
Janiczek: I think we can get rid of this

src/compiler/Stage/Optimize.elm
```diff
@@ -46,6 +47,26 @@ optimizePlus located =
...
+
+
+optimizeCons : Typed.LocatedExpr -> Maybe Typed.LocatedExpr
+optimizeCons located =
+    case Typed.getExpr located of
+        Typed.Cons l r ->
+            case Typed.getExpr r of
+                Typed.List list ->
+                    Just (Typed.mapExpr (\_ -> Typed.List (l :: list)) r)
```
Janiczek: I see you added `Typed.mapExpr`, and the optimization looks pretty nice with it. I know it's out of the scope of the PR, but can you please try doing the same for the `optimizePlus` function? It will be good to have them consistent.

src/compiler/Stage/Emit/JavaScript.elm
```diff
@@ -45,6 +45,9 @@ emitExpr located =
         Plus e1 e2 ->
             "(" ++ emitExpr e1 ++ " + " ++ emitExpr e2 ++ ")"
 
+        Cons e1 e2 ->
+            "[].concat([" ++ emitExpr e1 ++ "]," ++ emitExpr e2 ++ ")"
```
Janiczek: I know it's inconsequential, but it just dawned on me that it can be `[e1].concat(e2)` instead of `[].concat([e1],e2)`. WDYT?

tests/ParserTest.elm
```diff
@@ -912,6 +912,28 @@ expr =
                   )
                 ]
               )
+            , ( "cons"
+              , [ ( "simple case"
+                  , "1 :: []"
```
Janiczek: Let's add tests for whether it works with different amount of spaces (none, multiple). Don't do newlines for now

tests/TestHelpers.elm
```diff
@@ -60,3 +61,14 @@ typedString str =
...
+
+typedIntList : List Int -> Typed.LocatedExpr
+typedIntList list =
+    located
+        ( Typed.List
+            (list
+                |> List.map typedInt
+            )
+        , Type.List Type.Int
+        )
```
Janiczek: Just a style nitpick:
```elm
typedIntList : List Int -> Typed.LocatedExpr
typedIntList list =
    located
        ( Typed.List (List.map typedInt list)
        , Type.List Type.Int
        )
```

halfzebra:
Thanks for the feedback! I'll address the comments asap 👍

src/compiler/Stage/Emit/JavaScript.elm
halfzebra: Good idea, I've implemented this 👍

src/compiler/Stage/Optimize.elm
halfzebra: It's done! 👍

tests/ParserTest.elm
halfzebra: This is also done! 🎉

halfzebra:
Thanks a lot for helping with this PR, proper error substitution makes a lot of sense!

Please let me know if this needs more attention!

PR #56, opened 2019-08-09
Refactor modules

Janiczek:


harrysarson:
Our of curiosity: what is the difference between the old dictionary library and the new one?

Janiczek:
@harrysarson any-dict is normal Dict + a function from your key to comparable. assoc-list is List-based, so no comparable requirement but different performance characteristics

harrysarson:
Makes sense!

PR #59, opened 2019-08-17
A rough outline of cmds in elm

harrysarson:
Plus thoughts on implementing them in elm.

### DIFF
--- a/docs/core/cmds-in-elm.md
+++ b/docs/core/cmds-in-elm.md
@@ -1,0 +1,93 @@
+# Implementing the `Cmd` type in elm
+
+`Cmd`s are a fundamental type in elm.
+However, what _is_ an elm `Cmd`?
+The [official docs](https://package.elm-lang.org/packages/elm/core/1.0.2/Platform-Cmd) say
+
+> A command is a way of telling Elm, “Hey, I want you to do this thing!” So if you want to send an HTTP request, you would need to command Elm to do it. Or if you wanted to ask for geolocation, you would need to command Elm to go get it.
+>
+> Every Cmd specifies (1) which effects you need access to and (2) the type of messages that will come back into your application.4
+
+## The ananatomy of a `Cmd`
+
+A `Cmd` variable in an elm program will have one of the following JavaScript structures.
+Note that in the generated JavaScript the discriminants (`__2_NODE`, etc.) are replaced to reduce the size of the file.
+
+1. A `Cmd` created in elm using ```batch : List (Cmd a) -> Cmd a```.
+
+    `list` is generated JavaScript for an elm `List (Cmd a)`.
+
+    ```js
+    {
+        $: __2_NODE,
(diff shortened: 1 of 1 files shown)


harrysarson:
I have the beggining of an idea for _almost_ elm only implementation of `Cmd`s.

It involves a the following:

```elm
changeType : a -> b
changeType = 
    Elm.Kernel.changeType


type Opaque = Opaque


type Cmd a
    = Batch (List Cmd a)
    | Leaf String Opaque
    | Map (Opaque -> a) (Cmd Opaque)


map : (a -> b) -> Cmd a -> Cmd b
map fn cmda =
    Map (changeType fn) (changeType cmda)
```

In kernel land:

```js
function __kernel_whatsit_changeType(x) { return x; }
```

---

The root problem is that the elm type system cannot express this kind of data. (Aside: look at [the `and_then()` method](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.18/futures/future/trait.TryFutureExt.html#method.and_then) and [the `AndThen`  struct](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.18/futures/future/struct.AndThen.html) for how rust uses `traits` to hide the extra type parameters). This comment proposes adding a function that lies about the type signature allowing us to fudge these more complicated types.

I hear what you are rightly saying about the risks of this (some very strange type errors when elm attempts to interpret an integer as a list).

I am writing this as a comment rather than ammending the document because the idea is very rough but I am about to go on holiday and wanted to get my idea down somewhere before I leave.

docs/core/cmds-in-elm.md
```diff
@@ -0,0 +1,93 @@
...
+    | Leaf String Problem1
+    | Map (Problem2 -> a) (Cmd Problem 2)
+```
+
+However, there are two problems
+
+1. The type parameter `a` in `Cmd a` describes the type of message the runtime can send to an elm problem **not** the type of the payload that a `Cmd` has.
+    As a concrete example, calling `port xxx : String -> Cmd never` creates a `Cmd never` which contains a `String` payload but `String` does **not** appear in the type signature.
+
+    Elm's type system does not allow a type to contain a parameterised type unless the paramterised type is a parameter of the type.
```
turboMaCk, line 79: I believe this is in fact backed by Json Encoder and `Value` type which eliminates (which is derived for some types) which eliminates the need for existential type. That's being said it's just a gut feeling - I might be wrong.

harrysarson: What is written here could with a big update. My understanding of the effects system in elm has increased a lot since I wrote it.

Saying that I still think that some hidden parameterised type is needed. I don't think I have a good argument to hand though. One to maybe come up with at some point.

harrysarson:
closing as this isn't looking like it is a path to merging. These sort of docs probably belong in another-elm's repository instead too.

Issue #42, opened 2019-07-11
Add DesugarTests


Janiczek:
For what should be in them: don't check that it "does nothing". Check the interesting stuff that happens in the `desugarExpr` function. Eg.:

* [ ] Var: finds the module according to the surrounding context (imports and other top level definitions)
* [ ] Lambda: gets curried, ie. `\a b -> c` becomes `\a -> \b -> c`
* [ ] List: keeps the original order

PR #63, opened 2019-10-16
add test for desugar frontend lambdas

sebsheep:


tests/DesugarTest.elm
```diff
@@ -0,0 +1,76 @@
+module DesugarTest exposing (desugarTest)
+
+--import Elm.Data.Type as Type exposing (Type(..))
```
Janiczek: we can remove unused code

Janiczek:
Looks good so far! Are you planning to continue with the tests written out in #42?

sebsheep:
Yes I will. I had a look on
https://github.com/elm-in-elm/compiler/blob/master/src/Stage/Desugar.elm#L237
and I think the issue is deeper than what is described in the todo : the
elm compiler raises an error in case of ambiguity, so the previous function
should return smthg like `Result (List (Maybe ModuleName)) (Maybe
ModuleName)` (but it seems like we have to do a lot of refactor to get a
clean solution ; maybe we can postpone this "todo" for later).

Can write tests for it and currently ignore "ambiguous cases".

Le mer. 16 oct. 2019 à 21:33, Martin Janiczek <notifications@github.com> a
écrit :

Janiczek:
Yes, I'll create a separate GitHub issue for throwing "ambiguous" error instead of `Dict.find`-ing the first module that applies. A part of that issue will be adding this kind of desugaring test. Let's skip it here then.

Janiczek:
@sebsheep I have implemented the "Ambigous Name" check on the `master` branch. You can add tests for that one too then :smile:

See and `make` the current example-project in `master` for an example of the error firing.

Janiczek:
If I'm correct, this has been all done in #64 and can be closed?

PR #64, opened 2019-10-27
Test desugar

sebsheep:
Here is my attempt to test the name resolution with multiple modules.

All the tests pass! Tell me if you see situation I don't test you'd like to test.

### DIFF
--- a/tests/DesugarTest.elm
+++ b/tests/DesugarTest.elm
@@ -1,6 +1,6 @@
 module DesugarTest exposing (desugarTest)
 
-import Dict exposing (Dict)
+import Dict
 import Elm.AST.Canonical as Canonical
 import Elm.AST.Canonical.Unwrapped as CanonicalU
 import Elm.AST.Frontend as Frontend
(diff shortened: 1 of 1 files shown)


Janiczek:
Looks pretty solid!

I wonder if an inline style wouldn't be more readable, but it would probably still need some helper functions because of the `Modules` dict in the API... :thinking: 

But what I think could work and be nicer is making test cases from a table. Because on a high level, I've been thinking about it like this:

| module contains | imported contains | import | asks for | RESULT |
|---|---|---|---|---|
|  |  |  | x | Var not found |
| x |  |  | x | This.x |
| |  | NonExisting | x | Module not found |
|  |  | Imported | x | Var not found |
|  | x | Imported | x | Var not found |
|  | x | Imported | Imported.x | Imported.x |
|  | | Imported | Imported.x | Var not found |
|  | x | Imported as I | x | Var not found |
|  | x | Imported as I | Imported.x | Module not found |
|  | x | Imported as I | I.x | Imported.x |
|  | x | Imported exposing (y) | x | Var not found |
|  | x | Imported exposing (x) | x | Imported.x |
|  | x | Imported exposing (x) | Imported.x | Imported.x |

etc.

It doesn't need to be exhaustive, but maybe something like

```elm
type alias TestCase =
  { moduleContains : List Declaration
  , imports : List Import
  , importedModules : List (ModuleName, Exposing, List Declaration)
  , inputVar : (Maybe ModuleName, VarName)
  , expectedResult : Result DesugarError (ModuleName, VarName)
  }

toTest : TestCase -> Test

tests = List.map toTest [... testCases ...]
```
might look nicer? (I honestly don't know, it might very well be too much boilerplate for little gain... but it would map nicer to that table view of the behaviour I outlined above.)

Please tell me if you'd like to try this refactor. If not, let's merge as is :)

sebsheep:
Yeah, I didn't think about use types to write tests.... It would be indeed
a great readability improvement!

However, I'd go further by defining some custom types directly bounded to
the tests (build a complete Import or Declaration statement is kind of a
hassle, that's why i've defined so much helpers :) ):

{ thisModuleName = "Main"
, moduleVars = ["a", "b"]
, imports = [BareImport "A", ImportExposing "B" ["a"], ImportAliased "C"
"D"]
, availableModules = [ {name = "A", vars = ["a", "c", "d"]}, {name="B",
vars=["a", "e", "f"]}]
, inputVar = (Just "A" , "a")
, expectedResult = Ok ( "Main", "a")
}

I'll try to update the code today (no guarantee :( )

Le dim. 27 oct. 2019 à 23:46, Martin Janiczek <notifications@github.com> a
écrit :

sebsheep:
So here is my new version from your precious advice! I feel a lot happier with code!

Janiczek:
The tests look much better to me :) Thanks!

PR #65, opened 2019-11-09
Add type annotations support

Janiczek:
* [x] **parser tests**
* [x] **parser**
* [x] **desugaring** - check var and annotation name are the same, change from TypeAnnotation to Type
* [x] **desugar tests**
* [x] **type checking**: unify the type in annotation with the type of the expression, change from Type to Never
  * the changing to Never might be a bad idea sine that loses the type annotation information for any emitting... I guess we'll have to encounter the need for it and only then stop dropping the data
* [x] optimizations? at least think about it
* [x] emit? at least think about it

TODO

* [x] check the currentness of doc comments near the changed types and aliases
* [x] think about the new Function {to,from} naming (instead of the old {arg, result}) ... is it really better?
* [ ] check multiple SubstitutionMaps being created and merged together...
* [x] type parsers contexts

The plan:
1. parse type annotation into `Declaration`
2. when desugaring, check the names in `decl.typeAnnotation` and `decl.name` are the same (and drop the type annotation name)
3. when inferring types, check that the types unify, restrict the final type! (and drop the type annotation altogether)


src/Stage/Parse/Parser.elm
```diff
@@ -868,3 +872,28 @@ oneOrMoreHelp spaces p vs =
         , P.succeed ()
             |> P.map (always (P.Done (List.reverse vs)))
         ]
+
+
+typeAnnotation : Parser_ TypeAnnotation
+typeAnnotation =
+    -- TODO don't support newline without a space afterward... see the commented out tests
```
Janiczek: don't forget

tests/ParserTest.elm
```diff
@@ -1010,3 +1014,92 @@ contextToString context =
...
+            --  )
+            -- , ("float", "Float", Type.Float)
+            -- , ("char", "Char", Type.Char)
+            -- , ("string", "String", Type.String)
+            -- , ("bool", "Bool", Type.Bool)
+            -- TODO list
+            -- TODO tuple
+            -- TODO tuple3
+            -- TODO parametric type
+            -- TODO record type and all the rest of the possible types...
```
Janiczek: don't forget all these commented tests

tests/ParserTest.elm
```diff
@@ -1010,3 +1014,92 @@ contextToString context =
...
+            (List.map runTest
+                [ ( "canonical format", "x : Int", Just example )
+                , ( "no spaces", "x:Int", Just example )
+                , ( "multiple spaces before", "x   : Int", Just example )
+                , ( "multiple spaces after", "x :   Int", Just example )
+                , ( "newline and space before", "x\n  : Int", Just example )
+                , ( "newline and space after", "x :\n Int", Just example )
+
+                --, ( "newline before", "x\n: Int", Nothing )
+                --, ( "newline after", "x :\nInt", Nothing )
```
Janiczek: don't forget

src/Elm/Data/Declaration.elm
```diff
@@ -11,12 +11,14 @@ module Elm.Data.Declaration exposing
...
 
 import Elm.Data.Type exposing (Type, TypeArgument)
+import Elm.Data.TypeAnnotation exposing (TypeAnnotation)
 import Elm.Data.VarName exposing (VarName)
 
 
 {-| -}
 type alias Declaration expr =
     { module_ : String
+    , typeAnnotation : Maybe TypeAnnotation -- TODO desugar from Maybe TypeAnnotation to Maybe Type
```
Janiczek: don't forget

src/Stage/Parse/Parser.elm
```diff
@@ -325,12 +325,22 @@ declarations =
...
+        (\typeAnnotation_ name body module__ ->
             { module_ = module__
+            , typeAnnotation = typeAnnotation_
             , name = name
             , body = body
             }
         )
+        |= P.oneOf
+            -- TODO refactor the `backtrackable` away
+            -- TODO is it even working correctly?
```
Janiczek: don't forget

harrysarson:
> 3. when inferring types, check the types are the same (and drop the type annotation type)

The types might not be the same (?) The annotations might be more restrictive than the inferred type I think.

Janiczek:
@harrysarson Thanks! Yeah what I meant was probably that they _unify_. Will fix the note above.

src/Stage/Desugar.elm
```diff
@@ -180,6 +186,48 @@ desugarExpr modules thisModule located =
...
+    x =
+        123
+
+If they don't match, throw an error:
+
+     x : Int
+     y =
+         123
+
+TODO test
```
Janiczek: don't forget

Janiczek:
Just a small status update:

I believe I'm very close to finally finishing this.

The main pain of this PR is that we have to thread `SubstitutionMap` through all `unify` calls. No longer does `unify` take no SubstitutionMap and return a new one... That changes types at a lot of places...

sebsheep:
Oh my....!!!

Janiczek:
The `Var Id -> Var String` thing is quite big. Didn't get through all of it yesterday, hoping to finish tonight. We'll see :)

yarn.lock
```diff
@@ -0,0 +1,2853 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
harrysarson, line 2: I am sure you will catch this file but I thought pointing it out would not hurt.

I think it was committed by accident!

Janiczek: @harrysarson I did this on purpose. Is it controversial?

harrysarson: Okay makes sense! Ignoe me then

integration-tests/desugar/type-annotation-check-varname-err-not-same/test.test.js
```diff
@@ -0,0 +1,10 @@
+const test = require('ava');
+const {exec} = require('../../setup');
+
+test.failing(
+    "desugar: type annotation var name doesn't agree with the declaration var name",
+    exec,
+    __dirname,
+    ["-m", "src/Main.elm"],
+    async program => ({snapshot: await program})
```
harrysarson, line 9: On the back of #105 I think something more like the test in https://github.com/elm-in-elm/compiler/blob/1c8669fb0e7907ed00cd2f6d854803f2f67825a9/integration-tests/errors/missing-equals/test.test.js would avoid the need for `.failing` here.

```suggestion
    async (program, t) => {
        const snapshot = await t.throwsAsync(program);
        t.is(snapshot.code, 1);
        return snapshot;
    }
```

Janiczek: Thanks Harry! I just wanted to ping you here :)

I'll try to incorporate that suggestion to all the  *-err-* tests here when I get to my PC.

harrysarson: I think with the magic I have put in the [exec macro](https://github.com/elm-in-elm/compiler/blob/1c8669fb0e7907ed00cd2f6d854803f2f67825a9/integration-tests/setup.js) I have made the mistake of reducing boiler plate code at the cost of code readability.

Maybe I should look at making this macro a bit less magic. (For reference test macros are described [here](https://github.com/avajs/ava/blob/19c4f35fbae740268dedeeb4d450f813f2d5e85e/docs/01-writing-tests.md#reusing-test-logic-through-macros)).

Janiczek:
@harrysarson does this commit look OK to you?

PR #66, opened 2019-11-10
Record parser

sebsheep:
Work in progress. PR created to get feedback along the way.

### DIFF
--- a/src/Stage/Emit/JavaScript.elm
+++ b/src/Stage/Emit/JavaScript.elm
@@ -140,7 +140,7 @@ emitExpr located =
                 bindingsJS =
                     bindings
                         |> Dict.values
-                        |> List.map (\binding -> mangleVarName binding.name ++ " : " ++ emitExpr binding.body)
+                        |> List.map (\binding -> mangleVarName binding.name ++ ": " ++ emitExpr binding.body)
                         |> String.join ", "
             in
             "{" ++ bindingsJS ++ "}"
--- a/src/Stage/Emit/JavaScript.elm
+++ b/src/Stage/Emit/JavaScript.elm
@@ -140,7 +140,7 @@ emitExpr located =
                 bindingsJS =
                     bindings
                         |> Dict.values
-                        |> List.map (\binding -> mangleVarName binding.name ++ " = " ++ emitExpr binding.body)
+                        |> List.map (\binding -> mangleVarName binding.name ++ " : " ++ emitExpr binding.body)
                         |> String.join ", "
             in
             "{" ++ bindingsJS ++ "}"
--- a/src/Stage/Emit/JavaScript.elm
+++ b/src/Stage/Emit/JavaScript.elm
@@ -135,6 +135,16 @@ emitExpr located =
         Tuple3 e1 e2 e3 ->
             "[" ++ emitExpr e1 ++ "," ++ emitExpr e2 ++ "," ++ emitExpr e3 ++ "]"
 
+        Record bindings ->
+            let
+                bindingsJS =
+                    bindings
+                        |> Dict.values
+                        |> List.map (\binding -> mangleVarName binding.name ++ " = " ++ emitExpr binding.body)
+                        |> String.join ", "
+            in
+            "{" ++ bindingsJS ++ "}"
+
 
 emitDeclaration : Declaration Typed.LocatedExpr -> String
 emitDeclaration { module_, name, body } =
--- a/src/Stage/Parse/Parser.elm
+++ b/src/Stage/Parse/Parser.elm
@@ -351,6 +351,7 @@ expr =
             , tuple
             , tuple3
             , parenthesizedExpr
+            , record
             ]
         , andThenOneOf =
             -- TODO test this: does `x =\n  call 1\n+ something` work? (it shouldn't: no space before '+')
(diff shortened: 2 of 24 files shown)


Janiczek:
Looks great so far!

src/Stage/Parse/Parser.elm
```diff
@@ -804,6 +805,21 @@ tuple3 config =
...
 
 
+record : ExprConfig -> Parser_ LocatedExpr
+record config =
+    P.succeed Frontend.Record
+        |= P.sequence
+            { start = P.Token "{" ExpectingRecordLeftBrace
+            , separator = P.Token "," ExpectingRecordSeparator
+            , end = P.Token "}" ExpectingRecordRightBrace
+            , spaces = spacesOnly
```
Janiczek: I'm wondering about this vs `P.spaces` (but there is quite possibly some trickery with needing the same indentation etc., which is problematic in more places already)

src/Stage/Emit/JavaScript.elm
```diff
@@ -135,6 +135,16 @@ emitExpr located =
         Tuple3 e1 e2 e3 ->
             "[" ++ emitExpr e1 ++ "," ++ emitExpr e2 ++ "," ++ emitExpr e3 ++ "]"
 
+        Record bindings ->
+            let
+                bindingsJS =
+                    bindings
+                        |> Dict.values
+                        |> List.map (\binding -> mangleVarName binding.name ++ " = " ++ emitExpr binding.body)
+                        |> String.join ", "
```
Janiczek: I'm pretty sure this is not JavaScript syntax :D

sebsheep: Oh my! Too much elm in my life!

src/Stage/Parse/Parser.elm
Janiczek: We don't have to solve this in this PR - we can do a sweep of "what to do with multi-line expressions" for all expressions in another PR later.

sebsheep: yeah, I've just taken the list parser and "did the same", aka "only support one-line data structure" for the moment.

src/Stage/Emit/JavaScript.elm
Janiczek: You're saying that like it's a bad thing :stuck_out_tongue:

Janiczek:
So... it seems like the only things missing currently are parser tests and emit tests ... maybe some type-inference tests

sebsheep:
Well, the fuzzy tests don't work. I don't really understand the error I have, and if it is because of my tests or because of my code.

E.g. here is one of the mismatch:
```
Given If { else_ = Record (Dict.fromList [("a",{ body = If { else_ = Int -1042664441, test = Bool True, then_ = Int -658793470 }, name = "a" })]), test = Bool False, then_ = Record (Dict.fromList [("a",{ body = If { else_ = Int -932932175, test = Bool False, then_ = Int -1489278948 }, name = "a" })]) }

    Err (TypeMismatch (Record (Dict.fromList [("a",Int)])) (Record (Dict.fromList [("a",Var 9)])))
    ╷
    │ Expect.equal
    ╵
    Ok (Record (Dict.fromList [("a",Int)]))
```

Where does this `Var 9` comes from?
 @Janiczek could you take a look?

tests/EmitTest.elm
```diff
@@ -250,6 +250,64 @@ javascript =
...
+                      , """[(1 + 41),"Hello"]"""
+                      )
+                    , ( "tuple and cons in record"
+                      , Record
+                            (Dict.fromList
+                                [ ( "a", { name = "a", body = typed (Tuple (typedInt 2) (typedInt 3)) } )
+                                , ( "b", { name = "b", body = typed (Cons (typedInt 2) (typedIntList [ 3, 4 ])) } )
+                                ]
+                            )
+                      , """{a : [2,3], b : [2].concat([3, 4])}"""
```
Janiczek: This is absolutely an inconsequential nitpick but I'd like to go with the JS convention of `a: 1`, not `a : 1` (no space before)

sebsheep: seems reasonable

sebsheep:
All is good! Ask for final review :)

Janiczek:
@sebsheep 
Please also update the table in README:
table

I believe the parser needs to have a :warning: because of the multi-line thing (note 3?)
And some of the columns might not apply - just put a :heavy_check_mark: there.

---

Oh and also, if you feel like it, add yourself to the contributors at the bottom of the README ;)

tests/InferTypesTest.elm
```diff
@@ -286,6 +302,23 @@ typeToString =
...
+                )
+            , runTest
+                ( "one field record"
+                , Record <| Dict.fromList [ ( "a", Int ) ]
+                , "{ a : Int }"
+                )
+            , runTest
+                ( "two fields record"
+                , Record <| Dict.fromList [ ( "a", Int ), ( "b", String ) ]
+                , "{ a : Int, b : String }"
```
Janiczek: Can we just add one small test here about parametric types, to be sure?
`Record <| Dict.fromList [("foo", Var 0)]` :arrow_right: `{ foo : a }` or something like that.

Janiczek: I've checked the rest of the PR so after this addition we can merge :tada: :slightly_smiling_face:

Janiczek:
Thanks @sebsheep !

PR #68, opened 2019-11-12
make the InferTypes.unify case more explicit

sebsheep:
Problem: when a newcomer contributes to the project
for the first time, the comment in Elm/Data/Type.elm asking for adding
a case in the InfertTypes.unify function is meaningless
since the developer does not understand what this comment
is talking about. Thus, the developer could skip this comment,
leading to hard-to-find bug instead of easily understandable
compiler error message.

Indeed, the unify function pattern matchs against the Joker pattern "_"
at the end, silently ignoring new types.

Solution: Remove Joker pattern and pattern match against
    (T a b, T c d)

AND
    (T _ _, _)
for all types T.

There still is a comment, but this one is more local than the previous one.
Moreover this comment just asks "please do the same" instead of
"please do a specific thing".

### DIFF
--- a/src/Stage/InferTypes/Unify.elm
+++ b/src/Stage/InferTypes/Unify.elm
@@ -30,29 +30,89 @@ unify t1 t2 substitutionMap =
 
     else
         case ( t1, t2 ) of
+            {- Do not use joker pattern "_" at the end of this case and
+                pattern match all the types couples with
+
+                 ( T a b, T c d) ->
+                     ...
+
+                 ( T _ _, _) ->
+                     Err ( TypeMismatch t1 t2, substitutionMap )
+
+               It is definitively verbose but prevent the developer to skip this important
+               function: the compiler will throw an "MISSING PATTERNS" in this case.
+            -}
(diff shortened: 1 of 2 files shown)


src/Stage/InferTypes/Unify.elm
```diff
@@ -30,30 +30,90 @@ unify t1 t2 substitutionMap =
...
+                pattern match all the types couples with
+
+                 ( T a b, T c d) ->
+                     ...
+
+                 ( T _ _, _) ->
+                     Err ( TypeMismatch t1 t2, substitutionMap )
+
+               It is definitively verbose but prevent the developer to skip this important
+               function: the compiler will throw an "MISSING PATTERNS" in this case.
```
Janiczek: `prevent` :arrow_right: `prevents`
`"MISSING PATTERNS"` :arrow_right: `"MISSING PATTERNS" error`

Janiczek:
Much better without the wildcard, thanks!
After the comment change we can merge.

harrysarson:
I like!

PR #69, opened 2019-11-12
Add checks for duplicated fields in records

miniBill:
This is a work in progress: it's missing a nice error message and tests. Soliciting feedback.

### DIFF
--- a/src/Elm/Compiler/Error.elm
+++ b/src/Elm/Compiler/Error.elm
@@ -150,6 +150,7 @@ type DesugarError
         }
     | DuplicateRecordField
         { name : VarName
+        , insideModule : ModuleName
         , firstOccurrence : Located ()
         , secondOccurrence : Located ()
         }
--- a/src/Elm/Compiler/Error.elm
+++ b/src/Elm/Compiler/Error.elm
@@ -232,14 +232,18 @@ toString error =
                         ++ "\n\nChange your imports to resolve this ambiguity!"
 
                 DuplicateRecordField { name, firstOccurrence, secondOccurrence } ->
-                    "This record has multiple `"
-                        ++ name
-                        ++ "` fields. One here:\n\n"
-                        --TODO toString for Located
-                        ++ "TODO\n\n"
-                        ++ "And another one here:\n\n"
-                        ++ "TODO\n\n"
-                        ++ "How can I know which one you want? Rename one of them!"
+                    {- TODO
+                       This record has multiple `a` fields. One here:
+
+                       12|     { a = 123
+                                 ^
+                       And another one here:
+
+                       13|     , a = ()
+                                 ^
+                       How can I know which one you want? Rename one of them!
+                    -}
+                    "This record has multiple `" ++ name ++ "` fields."
 
         TypeError typeError ->
             case typeError of
--- a/src/Elm/Compiler/Error.elm
+++ b/src/Elm/Compiler/Error.elm
@@ -17,6 +17,7 @@ module Elm.Compiler.Error exposing
 -}
 
 import Elm.Data.FilePath exposing (FilePath)
+import Elm.Data.Located exposing (Located)
 import Elm.Data.ModuleName exposing (ModuleName)
 import Elm.Data.Type as Type exposing (Type)
 import Elm.Data.Type.ToString as TypeToString
@@ -147,6 +148,11 @@ type DesugarError
         , insideModule : ModuleName
         , possibleModules : List ModuleName
         }
+    | DuplicateRecordField
+        { name : VarName
+        , firstOccurrence : Located ()
+        , secondOccurrence : Located ()
+        }
 
 
 {-| Errors encountered during [typechecking](Elm.Compiler#inferExpr).
@@ -224,6 +230,16 @@ toString error =
                                 possibleModules
                             )
                         ++ "\n\nChange your imports to resolve this ambiguity!"
+
+                DuplicateRecordField { name, firstOccurrence, secondOccurrence } ->
+                    "This record has multiple `"
+                        ++ name
+                        ++ "` fields. One here:\n\n"
+                        --TODO toString for Located
+                        ++ "TODO\n\n"
+                        ++ "And another one here:\n\n"
+                        ++ "TODO\n\n"
+                        ++ "How can I know which one you want? Rename one of them!"
 
         TypeError typeError ->
             case typeError of
--- a/src/Stage/Desugar.elm
+++ b/src/Stage/Desugar.elm
@@ -40,6 +40,12 @@ desugarExpr modules thisModule located =
         return expr =
             Ok (Located.replaceWith expr located)
 
+        andThen fn =
+            Result.andThen
+                (\expr ->
+                    fn expr |> map identity
+                )
+
         map fn =
             Result.map
                 (\expr ->
@@ -183,12 +189,28 @@ desugarExpr modules thisModule located =
             bindings
                 |> List.map (Binding.map recurse >> Binding.combine)
                 |> Result.combine
-                |> map
+                |> andThen
                     (\canonicalBindings ->
                         canonicalBindings
-                            |> List.map (\canonicalBinding -> ( canonicalBinding.name, canonicalBinding ))
-                            |> Dict.fromList
-                            |> Canonical.Record
+                            |> List.foldr
+                                (\canonicalBinding ->
+                                    Result.andThen
+                                        (\dict ->
+                                            case Dict.get canonicalBinding.name dict of
+                                                Just existing ->
+                                                    Err <|
+                                                        DuplicateRecordField
+                                                            { name = canonicalBinding.name
+                                                            , firstOccurrence = Located.replaceWith () existing.body
+                                                            , secondOccurrence = Located.replaceWith () canonicalBinding.body
+                                                            }
+
+                                                Nothing ->
+                                                    Ok <| Dict.insert canonicalBinding.name canonicalBinding dict
+                                        )
+                                )
+                                (Ok Dict.empty)
+                            |> Result.map Canonical.Record
                     )
 
 
--- a/src/Stage/Desugar.elm
+++ b/src/Stage/Desugar.elm
@@ -224,7 +224,7 @@ maybeDuplicateBindingsError moduleName bindings =
 
 {-| Find the first two elements in a list that duplicate a given property.
 
-Benchmarks were made <https://gist.github.com/xarvh/2e3ddf2e3f0f2c6e1f84505e1bff1f1b>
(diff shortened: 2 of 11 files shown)


src/Elm/Compiler/Error.elm
```diff
@@ -147,6 +148,11 @@ type DesugarError
         , insideModule : ModuleName
         , possibleModules : List ModuleName
         }
+    | DuplicateRecordField
+        { name : VarName
+        , firstOccurrence : Located ()
+        , secondOccurrence : Located ()
+        }
```
sebsheep: Should contain `insideModule` field.

sebsheep: You also should add test(s) in DesugarTest.elm

Janiczek:
About the `insideModule` - it will be needed to be able to print the module name, yes, but right now the rest of the errors are also kinda lax, so I'm willing to let that detail slip and only have a specific DesugarError for that specific scenario, we can go through errors and make them all nice later when we have some vision of what they should all contain etc.

miniBill:
Added a test

src/Stage/Desugar.elm
```diff
@@ -183,12 +189,28 @@ desugarExpr modules thisModule located =
...
+                                        (\dict ->
+                                            case Dict.get canonicalBinding.name dict of
+                                                Just existing ->
+                                                    Err <|
+                                                        DuplicateRecordField
+                                                            { name = canonicalBinding.name
+                                                            , firstOccurrence = Located.replaceWith () existing.body
+                                                            , secondOccurrence = Located.replaceWith () canonicalBinding.body
+                                                            }
+
```
showell: Is it worth generalizing this to something like `findDups .name .body (Located.replaceWith ()) DuplicateRecordField lst`?

miniBill: Honestly, I don't think so. I don't think it's usual at all to have more than 2 duplicate fields.

showell: Not sure I understand your comment.  I'm not saying that we should detect more than 2 duplicates.  Do you mean "not sure it's common to do duplicate detection anywhere else in the codebase"?

showell: My suggestion may have been unclear--I was thinking just have a helper that detects the first two dups generically.  But I'm not sure we need it yet.  The only other place that seems to do something like duplicate detection does something slightly different (ambiguous module names).

I feel like @Janiczek has some fancy module that collapses a dict into a dict of k -> dups.

Janiczek: Something like Dict.Extra.groupBy will surely exist. I wouldn't generalize this yet; we can look at the similar usecases later and see if they're common enough to be generalized. (in another PR possibly)

showell: I propose `ensureUnique`: 

``` elm
 65         bindings =
 66             [ { name = "fred", body = "line 3" }
 67             , { name = "mary", body = "line 5" }
 68             , { name = "fred", body = "line 7" }
 69             ]
 70
 71         ensureUnique : List ( comparable, v ) -> Maybe ( comparable, v, v )
 72         ensureUnique list =
 73             let
 74                 uniqueHelper : List ( comparable, v ) -> Maybe ( comparable, v, v )
 75                 uniqueHelper lst =
 76                     case lst of
 77                         [] ->
 78                             Nothing
 79
 80                         [ x ] ->
 81                             Nothing
 82
 83                         ( k1, v1 ) :: ( k2, v2 ) :: rest ->
 84                             if k1 == k2 then
 85                                 Just ( k1, v1, v2 )
 86
 87                             else
 88                                 ( k2, v2 ) :: rest |> uniqueHelper
 89             in
 90             list
 91                 |> List.sortBy Tuple.first
 92                 |> uniqueHelper
 93
 94         uncheckedBindings =
 95             bindings
 96                 |> List.map (\r -> ( r.name, r.body ))
 97
 98         result =
 99             case ensureUnique uncheckedBindings of
100                 Nothing ->
101                     Ok (Dict.fromList uncheckedBindings)
102
103                 Just ( name, body1, body2 ) ->
104                     Err { name = name, body1 = body1, body2 = body2 }
```

Janiczek: I'm on mobile but here's what I had in mind
```elm
bindings : List (Binding expr)

bindings
|> Dict.Extra.groupBy .name
|> Dict.Extra.find (\_ occurences -> List.length occurrences > 1)
|> Maybe.map (\name occurrences -> ... the error ...)
|> Maybe.withDefault ...
```

showell: Yeah, that seems the way to go!  I think it will make the code a bit easier to read, and it would also be easier to extend to some kind of more verbose error message if somebody had a bunch of duplication for some reason.

xarvh: I went for
```elm
findDuplicatesBy : (a -> comparable) -> List a -> Maybe ( a, a )
findDuplicatesBy property list =
    case list of
        [] ->
            Nothing

        head :: tail ->
            case List.Extra.find (\item -> property item == property head) tail of
                Just duplicate ->
                    Just ( head, duplicate )

                Nothing ->
                    findDuplicatesBy property tail
```

Which should be pretty fast and avoid allocations.

src/Elm/Compiler/Error.elm
```diff
@@ -1,10 +1,14 @@
...
+    exposing
+        ( DesugarError(..)
+        , EmitError(..)
+        , Error(..)
+        , ParseContext(..)
+        , ParseError(..)
+        , ParseProblem(..)
+        , TypeError(..)
+        , toString
+        )
```
xarvh: elm-format did this.
How much do we want to stick to elm-format?

Janiczek: Let's use elm-format. We can force the multiple items on the same row using `@docs foo, bar, baz` later.

tests/DesugarTest.elm
```diff
@@ -119,6 +121,32 @@ desugarTest =
                         , insideModule = "A"
                         }
             }
+        , test "desugar duplicate record field" <|
+            \_ ->
+                let
+                    x : Frontend.LocatedExpr
+                    x =
+                        frontendLambda "x" "x"
```
Janiczek: We probably don't need a lambda as the body - it should be irrelevant? Can we use `Frontend.Unit` instead perhaps?

tests/DesugarTest.elm
```diff
@@ -119,6 +121,32 @@ desugarTest =
...
+                in
+                located (Frontend.Record fields)
+                    |> Desugar.desugarExpr Dict.empty (moduleFromName "A")
+                    |> mapUnwrap
+                    |> Expect.equal
+                        (Err <|
+                            CompilerError.DuplicateRecordField
+                                { name = "a"
+                                , firstOccurrence = x |> Located.replaceWith ()
+                                , secondOccurrence = x |> Located.replaceWith ()
```
Janiczek: If we want to have the two occurence fields in the error even though we ignore them for now, perhaps we should check that they contain expected data? (You'd probably have to *not* use `located` helper for that particular test then.)

(Removing the occurence might be easier for now, and I wouldn't be against that :slightly_smiling_face: )

src/Stage/Desugar.elm
```diff
@@ -180,22 +181,67 @@ desugarExpr modules thisModule located =
...
+            )
+
+
+{-| Find the first two elements in a list that duplicate a given property.
+
+Implemented to avoid allocations.
+
+-}
+findDuplicatesBy : (a -> comparable) -> List a -> Maybe ( a, a )
+findDuplicatesBy property list =
```
Janiczek: O(n^2) complexity... I wonder if there's a better way. Not sure yet

sebsheep: Maybe be first ordering the list by field name? It would be O(nlog(n)).

Janiczek: https://ellie-app.com/7mJmZBnbQKCa1
:arrow_up: Possible alternative (I don't like it though) but I'm not sure it's better. Also I'm getting a bit lost in the complexity of Dict.get, filter, etc... It's O(n log n) too, right?
We should run a Benchmark on the two (three if we want to try the sorting) implementations!

xarvh: If performance is a problem, we might want to run a couple of benchmarks before thinking complexity.

sebsheep: Dict.get should be O(log(n)) (with n the number of (key,value) couples)

I don't see how Dict.filter could be less than O(n).

sebsheep: But yes as pointed out by @xarvh it is only a performance issue, not sure we should be really concerned by this right now since the records dont containt a huge number of fields most of the time, so the O(n^2) behavior should not be an issue (moreover in you Ellie @Janiczek you do a lot of map/filter/... copying the lists which can be worst than the O(n^2) if the n is not big).

That said, did you consider using List.Extra.unique* fonctions (the `allDifferent` function seems to be exactly what we need but the implementation isn't very efficient)?

https://github.com/elm-community/list-extra/blob/8.2.2/src/List/Extra.elm#L415-L456

xarvh: So, I got curious.

TL;DR: The O(n^2) `onn` algorithm is faster only up to 8 items. At 9 items, `sort` is faster.

I'm not sure if there is any reliable way to estimate memory footprint, my guesstimate is that `onn` won't allocate much and `sort` transforms the list into a js Array which is still O(n) and js will be optimized af so it's good.

I was hoping to get the best from both worlds and run one algorithm for long lists and the other for short lists, but estimating the list length is in itself expensive, so meh.


image

https://gist.github.com/xarvh/2e3ddf2e3f0f2c6e1f84505e1bff1f1b

showell: @xarvh Thanks for doing the benchmark!  I am not surprised that `sort` does better than o(n*n) for large numbers, of course, but I was surprised the threshold was so low.  My vote would be to use `sort` here, even though I understand the argument that most records are pretty small (and, probably, in fact fewer than 8 fields 99% of the time).  The `sort` code should be pretty easy to read and understand, and I don't like surprises for the worst case scenario.

This is obviously suggestive and just my opinion.  I think the n-squared approach is also fine.

xarvh: Agreed. Sort it is. We can always switch algorithm later on. I'll link the benchmark gist in the comments.

Janiczek: Thanks!

xarvh: Should be mergeable.

harrysarson: Can we commit this benchmark somewhere?

xarvh: Sure.

PR #70, opened 2019-11-13
[UP FOR GRABS] Parsing performance issue

Janiczek:
## THIS PERFORMANCE INVESTIGATION IS UP FOR GRABS

@showell Here's a branch with reproducible way to test the performance on your example.
Running `make` in the elm-in-elm root should be enough to get some numbers (if you have `bash`).
get-perf


Janiczek:
Running `make test-perf` on this branch rebased to current `master` yields these *parsing* times:
```
COMPILING THE COMPILER (NON-OPTIMIZED)

MEASURING! (NON-OPTIMIZED)
real	0m2.076s
user	0m2.338s
sys	0m0.034s

COMPILING THE COMPILER (OPTIMIZED)

MEASURING! (OPTIMIZED)
real	0m0.700s
user	0m0.897s
sys	0m0.050s
```

Janiczek:
Found the performance hog - parenthesis-related expressions and all their backtracking. `backtrackable` is really the root of all evil :)

I'm seeing 100ms times for that specific nested example expression now, instead of 1-2s.

PR #74, opened 2019-12-08
Simplify assign ids

sebsheep:
Problem: the `AssignIds.elm` uses the `IdSource` abstraction. This abstraction
guarantees that all ids are different by only exposing functions that increment
the id. However, the code using this abstraction needs to carefully deal with
the several `idSource` states, defeating the initial goal. Hence this abstraction
does not guarantee for what it was designed for, but add useless complexity
to the project.

Solution: remove the `IdSource` abstraction and use a bare `Int` instead
(aliased as `Id`).


Janiczek:
Alternative (which I failed writing a few times before): present a monadic interface that autoincrements and doesn't let you make an error (use a wrong IdSource etc.)...

I'll check this PR later today. Thank you!

sebsheep:
A few thoughts (mainly intuition):

1. I'd prefer "assumed explicit and simple state management" to "hard to work with and convoluted monadic trick". Of course, I'd be happy to have a "simple state management with monadic interface" :)
2. In case of explicit state management, I really think that "unused variable warning" from elm-analyze does good job by preventing us to mess up with the ids (at least, it prevented ME to do it); maybe we could add an "unused variable test" to the CI? 
3. The fact you failed writing this monadic interface maybe means there is no such simple solution or that we should entirely rethink the data structure (which could maybe lead to get rid of the mutual recursion)
4. Maybe the absence of monadic support in elm prevent us to write such a simple solution

Janiczek:
1. Same!
2. We can look at adding an elm-analyse step to the CI. The comment in this PR is also a good idea
3. I believe it's more about me not knowing how to write these monadic APIs than it being inherently impossible
4. I think all the monads in Elm (List, Random, Maybe, Result, Task,...) disprove this

sebsheep:
> 4. I think all the monads in Elm (List, Random, Maybe, Result, Task,...) disprove this

Working with a bunch of maybe value is not that handy :
```elm
f : Maybe a -> Maybe b -> Maybe c -> Maybe (a,b,c)
f a b c =
     a |>
      Maybe.andThen (\ innerA ->
              b |> Maybe.andThen (\ innerB ->
                        c |> Maybe.andThen (\ innerC ->
                                       Just (innerA, innerB, innerC)
 )))
```
compare to haskell's way:
```haskell
f a b c = do
    innerA <- a
    innerB <- b
    innerC <- c
    return (innerA, innerB, innerC)
```
That is what I meant.

Janiczek:
Thanks @sebsheep !

PR #78, opened 2019-12-25
Add emitter for AST data in JSON format

sgdan:
In order to try out Elm compilation with GraalVM and the Truffle API, I've added an emitter that produces JSON output containing the AST data. There's an "outputFormat" flag in the cli/index.js which can be switched to "JSON" to try it out.

See https://github.com/sgdan/truffle-elm where I've used the JSON output, it does something like this:
JSON AST
-> parse into Kotlin data classes
-> create Truffle nodes from AST data
-> run via GraalVM/Truffle interpreter or generate native linux executable

### DIFF
--- a/cli/index.js
+++ b/cli/index.js
@@ -1,10 +1,10 @@
 const fs = require('fs').promises; // needs Node.JS v10+
 
-const {Elm}          = require('../build/elm.js'); // build using Makefile... no Webpack around here!
-const {registerPort} = require('./utils.js');
+const { Elm } = require('../build/elm.js'); // build using Makefile... no Webpack around here!
+const { registerPort } = require('./utils.js');
 
 // Async/await is nice! (needs Node.JS v7.6+)
-(async function(){
+(async function () {
 
   console.log('---------------------------');
   console.log('-- STARTING THE COMPILER --');
@@ -15,7 +15,8 @@ const {registerPort} = require('./utils.js');
   const app = Elm.Main.init({
     flags: {
       mainFilePath: 'src/Main.elm',
-      elmJson: await fs.readFile(`${exampleProjectPath}/elm.json`, {encoding: 'utf8'}),
+      elmJson: await fs.readFile(`${exampleProjectPath}/elm.json`, { encoding: 'utf8' }),
+      outputFormat: "JavaScript" // alternative "JSON"
     }
   });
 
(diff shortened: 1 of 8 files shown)


cli/index.js
```diff
@@ -15,7 +15,8 @@ const {registerPort} = require('./utils.js');
   const app = Elm.Main.init({
     flags: {
       mainFilePath: 'src/Main.elm',
-      elmJson: await fs.readFile(`${exampleProjectPath}/elm.json`, {encoding: 'utf8'}),
+      elmJson: await fs.readFile(`${exampleProjectPath}/elm.json`, { encoding: 'utf8' }),
+      outputFormat: "JavaScript" // alternative "JSON"
```
Janiczek: So, hardcoded right now; no cmdline args reading, right?

src/Stage/Emit/JsonAST.elm
```diff
@@ -0,0 +1,221 @@
...
+                |> List.map emitDeclaration
+    in
+    Dict.singleton "out.json" ("[" ++ String.join "," declarations ++ "]")
+
+
+toJson : String -> List ( String, String ) -> String
+toJson astType fields =
+    (quote "type" ++ ":" ++ quote astType)
+        :: List.map (\( k, v ) -> quote k ++ ":" ++ v) fields
+        |> brace
```
Janiczek: I wonder if using Json.Encode wouldn't be a better solution here...

Janiczek:
Would be good to document that the JSON output is tied to the truffle-elm project.

Before merging this, I'd like to try Json.Encode instead of slapping strings together - or at least some rationale for why that approach was chosen...

Otherwise I think it's good enough and we can refactor and DRY the various emit phases a bit when we have more examples.

Cool project BTW, very exciting! :) Didn't expect native compilation that soon, and didn't know about Truffle / GraalVM enough to connect the dots!

sgdan:
No rush to merge, I'd like to look at the improvements you've suggested. I didn't even think of using Json.Encode, pretty sure that would be nicer. No good rationale for not using it!

Regarding the command line args, do you think those should be processed in `index.js` using something like `yargs` or handled somehow in the Elm code?

I suspect the truffle approach I took is more suitable for interpreted languages than compiled languages like Elm, but it's pretty simple to implement and might be useful for trying things out.

Janiczek:
I think the JS `yargs` solution is the saner one right now, you can explore that; but I'm even OK with keeping it hardcoded as it is for now.

Janiczek:
Hey @sgdan is it OK with you if I add you to the [contributors](https://github.com/elm-in-elm/compiler#contributors) table in README?

sgdan:
That would be great, thanks :)

PR #80, opened 2020-01-01
nice compiler error messages for parse problems

harrysarson:
I was playing around with the compiler but quickly got fustrated by the parser error messages.

Here is a quick solution that prints the offending line and points to the offending character.
The format of the output is based on the (very good) error messages of the rust compiler.

Here is an example output using the committed `example-project/src/Main.elm`:

```
Compiled in DEV mode. Follow the advice at https://elm-lang.org/0.19.1/optimize for better performance and smaller assets.
---------------------------
-- STARTING THE COMPILER --
---------------------------


---------------------------
-- COMPILER ERROR ---------
---------------------------
Parse problem: ExpectingEqualsSign
  --> src/Main.elm:4:3
   | 
4  | f x =
   |   ^ ExpectingEqualsSign
```

I implemented this by using the "assume all the information I want is availiable in `Elm.Compiler.Error.toString` and then keep fixing compiler errors untill it works" but I do not know this code base (at all really) so this may not be a great solution to the problem.

### DIFF
--- a/cli/elm.json
+++ b/cli/elm.json
@@ -16,6 +16,7 @@
             "elm-community/basics-extra": "4.0.0",
             "elm-community/dict-extra": "2.4.0",
             "elm-community/graph": "6.0.0",
+            "elm-community/list-extra": "8.2.2",
             "elm-community/maybe-extra": "5.1.0",
             "elm-community/result-extra": "2.3.0",
             "erlandsona/assoc-set": "1.1.0",
--- a/cli/elm.json
+++ b/cli/elm.json
@@ -16,7 +16,6 @@
             "elm-community/basics-extra": "4.0.0",
             "elm-community/dict-extra": "2.4.0",
             "elm-community/graph": "6.0.0",
-            "elm-community/list-extra": "8.2.2",
             "elm-community/maybe-extra": "5.1.0",
             "elm-community/result-extra": "2.3.0",
             "erlandsona/assoc-set": "1.1.0",
--- a/elm.json
+++ b/elm.json
@@ -38,6 +38,7 @@
         "elm-community/basics-extra": "4.0.0 <= v < 5.0.0",
         "elm-community/dict-extra": "2.4.0 <= v < 3.0.0",
         "elm-community/graph": "6.0.0 <= v < 7.0.0",
+        "elm-community/list-extra": "8.2.2 <= v < 9.0.0",
         "elm-community/maybe-extra": "5.0.0 <= v < 6.0.0",
         "elm-community/result-extra": "2.2.1 <= v < 3.0.0",
         "erlandsona/assoc-set": "1.0.0 <= v < 2.0.0",
--- a/elm.json
+++ b/elm.json
@@ -38,7 +38,6 @@
         "elm-community/basics-extra": "4.0.0 <= v < 5.0.0",
         "elm-community/dict-extra": "2.4.0 <= v < 3.0.0",
         "elm-community/graph": "6.0.0 <= v < 7.0.0",
-        "elm-community/list-extra": "8.2.2 <= v < 9.0.0",
         "elm-community/maybe-extra": "5.0.0 <= v < 6.0.0",
         "elm-community/result-extra": "2.2.1 <= v < 3.0.0",
         "erlandsona/assoc-set": "1.0.0 <= v < 2.0.0",
--- a/src/Elm/Compiler/Error.elm
+++ b/src/Elm/Compiler/Error.elm
@@ -207,13 +207,14 @@ toString error =
                         (List.map
                             (\{ problem, row, col, contextStack } ->
                                 let
-                                    filenameFromContext : List { a | context: ParseContext } -> Maybe FilePath
+                                    filenameFromContext : List { a | context : ParseContext } -> Maybe FilePath
                                     filenameFromContext contextStack_ =
                                         case contextStack_ of
                                             { context } :: rest ->
                                                 case context of
                                                     InFile name ->
                                                         Just name
+
                                                     _ ->
                                                         filenameFromContext rest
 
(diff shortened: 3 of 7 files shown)


src/Elm/Compiler/Error.elm
```diff
@@ -198,10 +202,56 @@ toString error =
...
+                                            |> List.filterMap
+                                                (\{ context } ->
+                                                    case context of
+                                                        InFile name ->
+                                                            Just name
+
+                                                        _ ->
+                                                            Nothing
+                                                )
+                                            |> List.Extra.last
```
Janiczek: The way this is set up, do we ever get more than one `InFile`? Couldn't we just use `List.Extra.find`?

harrysarson: Yes, I think we could. With fix

harrysarson:
Do you mind adding the list-extra dep?

Janiczek:
It's fine to have the list-extra dependency :)

src/Elm/Compiler/Error.elm
harrysarson: I am having second thoughts, currently we have:

```elm
filename =
    contextStack
        |> List.Extra.filterMap
            (\{ context } ->
                case context of
                    InFile name ->
                        Just name

                    _ ->
                        Nothing
            )
        |> List.Extra.last
````

I cannot think of a better way to write it using `List.Extra.find` than

```elm
filename =
    contextStack
        |> List.Extra.map
            (\{ context } ->
                case context of
                    InFile name ->
                        Just name

                    _ ->
                        Nothing
            )
        |> List.Extra.find 
            (\{ name_ } ->
                case name_ of
                    Just name ->
                        True

                    _ ->
                        False
            )
        |> Maybe.withDefault Nothing
```

which is horrible!

Thoughts?

Janiczek: True, when I tried, I got to similar issues. So what about custom recursion?
```elm
filename =
    findFirstJustcontextStack
    contextStack
        |> List.filterMap
            (\{ context } ->
                case c
```

Janiczek: True, when I tried I got into similar issues. So what about custom recursion? (Maybe this has some List.Extra equivalent...)
```elm
filename =
    findFirstJust 
        (\{ context } ->
            case context of
                InFile name ->
                    Just name
                _ ->
                    Nothing
        )
        contextStack

findFirstJust : (a -> Maybe b) -> List a -> Maybe b
findFirstJust fn list =
    case list of
        [] ->
            Nothing

        x :: xs ->
            case fn x of
                Just result ->
                    Just result

                Nothing ->
                    findFirstJust fn xs
```

Janiczek: (Or let's keep the `filterMap` and `last` / `first` / ..., considering the alternatives aren't much better after all)

harrysarson: How about:

```elm
filenameFromContext : List ParseContext -> Maybe FilePath
filenameFromContext contextList =
    case contextList of
        context :: rest ->
            case context of
                InFile name ->
                    Just name
                _ ->
                    filenameFromContext rest

        [] ->
            Nothing
```

src/Elm/Compiler/Error.elm
```diff
@@ -198,10 +202,56 @@ toString error =
...
+                                in
+                                "Parse problem: "
+                                    ++ parseProblemToString problem
+                                    ++ "\n  --> "
+                                    ++ (filename
```
harrysarson:
```suggestion
                                    ++ (filenameFromContext contextStack
```

src/Elm/Compiler/Error.elm
Janiczek: Looks good!

harrysarson: will sort next time I am on my laptop 👍

cli/elm.json
```diff
@@ -16,6 +16,7 @@
             "elm-community/basics-extra": "4.0.0",
             "elm-community/dict-extra": "2.4.0",
             "elm-community/graph": "6.0.0",
+            "elm-community/list-extra": "8.2.2",
```
harrysarson:
```suggestion
```

elm.json
```diff
@@ -38,6 +38,7 @@
         "elm-community/basics-extra": "4.0.0 <= v < 5.0.0",
         "elm-community/dict-extra": "2.4.0 <= v < 3.0.0",
         "elm-community/graph": "6.0.0 <= v < 7.0.0",
+        "elm-community/list-extra": "8.2.2 <= v < 9.0.0",
```
harrysarson:
```suggestion
```

harrysarson:
@Janiczek fixed! (I nolonger depend on list-exta either)

src/Elm/Compiler/Error.elm
```diff
@@ -16,13 +16,16 @@ module Elm.Compiler.Error exposing
...
+import Array
+import Elm.Data.FileContents exposing (FileContents)
 import Elm.Data.FilePath exposing (FilePath)
 import Elm.Data.Located exposing (Located)
 import Elm.Data.ModuleName exposing (ModuleName)
 import Elm.Data.Type as Type exposing (Type)
 import Elm.Data.Type.ToString as TypeToString
 import Elm.Data.VarName exposing (VarName)
 import Json.Decode as JD
+import List.Extra
```
Janiczek: this probably needs to get deleted for the build to pass

harrysarson:
```suggestion
```

harrysarson:
Do you want me to squash?

Janiczek:
github

Github can squash PRs automatically :)

Janiczek:
Great job, thank you @harrysarson

harrysarson:
My first propper (code) contribution to elm-in-elm!

PR #84, opened 2020-01-31
Cli tidying + integration tests

harrysarson:
1. Gives the CLI some love; removing the hardcoded "examples" directory in favour of using the current working directory.
2. Removes the default value for the main elm file. This might be something we want to add back in later but it seems (to me) prodent to not commit ourselves here. 
3. Adds integration tests using the [ava](https://github.com/avajs/ava) test runner.
4. The compiler exit code now reflects the compilation sucess/failure.


Janiczek:
Thanks! :heart:

I think these tests (comparing stderr and stdout) might be good for some kinds of tests (ie. "what error does this return?" - even though it might be even better to check the `Error` value in Elm with an unit test?)

But I don't want to depend on the `Debug.log` values in the success path of parsing etc. - what the one integration test currently does. These will go away and ... presumably our tests will start failing?

Perhaps I'm missing something about snapshot testing (never tried it). Does it have some solution for these situations?

---

One thing that might be *very* useful in these integration tests is running the emitted JS with `node` or something, and checking the result of running that code. WDYT?

harrysarson:
> even though it might be even better to check the Error value in Elm with an unit test?

It is better! These integration tests are meant to compliment, not replace unit tests. 

> But I don't want to depend on the Debug.log values in the success path of parsing etc. - what the one integration test currently does. These will go away and ... presumably our tests will start failing?

I think that the output (stdout/stderr) of a stable compiler should be stable. Obviously elm-in-elm is a _long_ way from being stable but these tests atleast allow us to track the output as it changes.

> Perhaps I'm missing something about snapshot testing (never tried it). Does it have some solution for these situations?

The idea is you can run `npx ava --update-snapshots` and instead of having failing tests the test runner will overwrite and update your snapshot files. As these files are tracked by git the changes to the output would be part of a pull request.

harrysarson:
I have added an extra feature to this pr: have the cli report compile failures by exiting with a non-zero exit code.

harrysarson:
Bump :)

Janiczek:
I guess let's give this a try. As said, I don't have experience with snapshot tests but what you explained about them makes me fear the maintenance cost a little bit less :D

harrysarson:
Ace! Let's see how it goes, I dont think this will be the final solution but my hope is it helps us get the ball moving and explore the space so we can better understand how we want these sort of integr

Issue #76, opened 2019-12-14
Make the let body see the let bindings


Janiczek:
Currently if you try to put
```elm
module Main exposing (main)

main =
    let
        x = 2
    in
        1 + x
```
into `example-project/src/Main.elm` and run it with `make`, it complains about `x` not being visible in the body. Make it so!

Depending on where the fix is, it might be a good idea to also add a regression unit test somewhere.

Feel free to discuss/ask how to solve this on our Discord!

alexshadley:
I'm taking a look at this right now! Will keep everyone up to date with my progress.

Janiczek:
Let's not forget the `test.failing(` :arrow_right: `test(` change in integration tests when this is done: https://github.com/elm-in-elm/compiler/pull/85#issue-375546599

harrysarson:
CI should fail if we do forget :)

gmilleramilar:
Looks like maybe this is fixed?  The contents of Main.elm in example-project are:

```
module Main exposing (main)

main =
    let
        x = 2
    in
        1 + x
```

Janiczek:
Yup seems like we fixed this as one of the last things before the project took a bit of a hiatus in 2020 😄 Closing :+1:

PR #85, opened 2020-02-14
Integration tests part 2

harrysarson:
A few further improvements to the integration tests landed in #84 plus adding three more integration tests.

The third is failing and will continue to fail untill a fix for #76 lands. The fix for ~76 will then need to include this patch:

```patch
diff --git a/integration-tests/typecheck/let-body-visibility/test.test.js b/integration-tests/typecheck/let-body-visibility/test.test.js
index 21f5d6e..3b239e6 100644
--- a/integration-tests/typecheck/let-body-visibility/test.test.js
+++ b/integration-tests/typecheck/let-body-visibility/test.test.js
@@ -1,7 +1,7 @@
 const test = require('ava');
 const {exec} = require('../../setup');
 
-test.failing(
+test(
     'typecheck: the let body can see the let bindings',
     exec,
     __dirname,
```


harrysarson:
@Janiczek ready for review!

Janiczek:
Hello @harrysarson sorry that this is taking so long. I'll try to get to it soon :crossed_fingers:

harrysarson:
Hello @Janiczek. Not a problem! Take care and have a nice afternoon :)

Makefile
```diff
@@ -1,3 +1,11 @@
+
+# elm-format does not support exlucding files so we explicitly list all the directories
```
Janiczek, line 2:
```suggestion
# elm-format does not support excluding files so we explicitly list all the directories
```

PR #104, opened 2020-05-15
Case of parser

pablohirafuji:
Hey, great project! Can't wait to see it working! Is this design of the `case of` parser ok?


pablohirafuji:
I know it's missing the `custom types`, but I can do that next, if I'm on the right track here.

Janiczek:
Hey @pablohirafuji cool stuff! I'll try to review this over the weekend. Thanks!

harrysarson:
Exciting

src/Stage/Emit/JavaScript.elm
```diff
@@ -128,6 +128,9 @@ emitExpr located =
             in
             "{" ++ bindingsJS ++ "}"
 
+        Case e branches ->
+            "TODO"
```
Janiczek, line 132: Do you want to figure out the emit phase in the same PR? We could talk through it here / on the Discord.

src/Stage/InferTypes/AssignIds.elm
```diff
@@ -279,3 +279,137 @@ assignIdsWithHelp currentId located =
...
+            assignPatternIdsWithHelp currentId (Located.unwrap locatedCanonicalPattern)
+    in
+    {- Keep location, for error context -}
+    ( Located.replaceWith typedPattern locatedCanonicalPattern
+    , newId
+    )
+
+
+assignPatternId : Id -> Typed.Pattern_ -> ( Typed.Pattern, Id )
+assignPatternId currentId located =
```
Janiczek, line 325: Seems like this is the same as `assignId`, just with different type. Can we reuse it, and make it polymorphic instead? Or are the types making it impossible?

src/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -385,6 +385,50 @@ generateEquations currentId located =
...
+                                ++ patternEquations
+                                ++ bodyEquations
+                            , nextId
+                            )
+                        )
+                        ( [], id1 )
+                        branches
+            in
+            ( testEquations
+                ++ branchesEquations
```
Janiczek, line 428:
```suggestion
            ( testEquations ++ branchesEquations
```

src/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -412,3 +456,165 @@ generateArgumentUsageEquations argumentId usages =
...
+        Typed.PVar _ ->
+            -- we can't make any assumptions here
+            ( [], currentId )
+
+        Typed.PRecord _ ->
+            -- we can't make any assumptions here
+            ( [], currentId )
+
+        Typed.PAlias pttrn_ _ ->
+            --TODO: not sure what to do here
```
Janiczek, line 481: If this is `(<SOME OTHER PATTERN> as foo)`, perhaps we should delegate to `generatePatternEquations currentId <SOME OTHER PATTERN>`?

If there are issues with `generatePatternEquations` taking `LocatedPattern` instead of `Pattern` or `Pattern_` then perhaps we should split this whole function into two, and use the "unwrapped" one when delegating in this `PAlias` pattern.

src/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -412,3 +456,165 @@ generateArgumentUsageEquations argumentId usages =
...
+            -- we can't make any assumptions here
+            ( [], currentId )
+
+        Typed.PVar _ ->
+            -- we can't make any assumptions here
+            ( [], currentId )
+
+        Typed.PRecord _ ->
+            -- we can't make any assumptions here
+            ( [], currentId )
```
Janiczek, line 478: I wonder if we can or can not. I guess we don't have enough "infrastructure" to be able to say "it's a record but I'm not gonna say more about what's inside"... I guess we can let this one be for now?

src/Stage/InferTypes/GenerateEquations.elm
```diff
@@ -412,3 +456,165 @@ generateArgumentUsageEquations argumentId usages =
...
+
+                ( leftEquations, id1 ) =
+                    generatePatternEquations currentId left
+
+                ( rightEquations, id2 ) =
+                    generatePatternEquations id1 right
+            in
+            ( -- For expression a :: [ b ]:
+              [ equals rightType (Type.List leftType) -- type of b is a List a
+              , equals type_ rightType -- a :: [ b ] is a List b
```
Janiczek, line 590: I get what you mena by these comments but they are confusing to me. I guess I'd skip the `[ b ]` and write `b` instead.

```suggestion
            ( -- For expression a :: b:
              [ equals rightType (Type.List leftType) -- type of `b` is `List a`
              , equals type_ rightType -- type of `a :: b` is `List a`
```
Still kinda weird (are `a` and `b` values or types?) but seems a little bit better to me.

src/Stage/Parse/Parser.elm
```diff
@@ -821,6 +830,227 @@ record config =
...
+        , andThenOneOf =
+            [ PP.infixRight 1 (P.symbol (P.Token "::" ExpectingConsOperator)) (Located.merge PCons)
+            , postfix 1
+                (P.succeed identity
+                    |. P.symbol (P.Token "as" ExpectingAsKeyword)
+                    |. ignorablesAndCheckIndent (<) ExpectingIndentation
+                    |= varName
+                    |> located
+                )
+                (Located.merge (\p v -> PAlias p (Located.unwrap v)))
```
Janiczek, line 886: Better var names please, I had trouble deciphering what `p` and `v` are.
```suggestion
                (Located.merge (\pattern alias -> PAlias pattern (Located.unwrap alias)))
```

src/Stage/Parse/Parser.elm
```diff
@@ -821,6 +830,227 @@ record config =
...
+        }
+        |> located
+        |> P.andThen
+            (\locatedPattern ->
+                case Located.unwrap locatedPattern of
+                    [ p1, p2, p3 ] ->
+                        Located.map (\_ -> PTuple3 p1 p2 p3)
+                            locatedPattern
+                            |> P.succeed
+                            |> P.inContext InTuple
```
Janiczek, line 1038:
```suggestion
                            |> P.inContext InTuple3
```

src/Stage/Parse/Parser.elm
```diff
@@ -821,6 +830,227 @@ record config =
...
+
+                    [ p1, p2 ] ->
+                        Located.map (\_ -> PTuple p1 p2)
+                            locatedPattern
+                            |> P.succeed
+                            |> P.inContext InTuple3
```
Janiczek, line 1044:
```suggestion
                            |> P.inContext InTuple
```

src/Stage/Parse/Parser.elm
```diff
@@ -835,6 +1065,68 @@ newlines =
     P.chompWhile ((==) '\n')
 
 
+checkIndent : (Int -> Int -> Bool) -> ParseProblem -> Parser_ ()
+checkIndent check error =
```
Janiczek, line 1069: Please let's leave some explanatory doc comment here. What are the two integers inside the `check` function (`intend` and `col`)? When does the parser fail and succeed? I feel like some examples of usage might be good for understanding too.

This is coming from somebody who didn't use `elm/parser`'s `indent` functionality yet (yeah...) - so that perhaps clears up why I want this documentation. Links to `elm/parser`'s documentation about `indent` and `col` are welcome!

src/Stage/Parse/Parser.elm
```diff
@@ -835,6 +1065,68 @@ newlines =
...
+            else
+                P.problem error
+        )
+        |= P.getIndent
+        |= P.getCol
+        |> P.andThen identity
+
+
+ignorablesAndCheckIndent : (Int -> Int -> Bool) -> ParseProblem -> Parser_ ()
+ignorablesAndCheckIndent check error =
```
Janiczek, line 1084: Similar to the above comment for `checkIndent` - and please tell us how does it differ from it (preferably with examples). Thanks!

src/Stage/Parse/Parser.elm
```diff
@@ -885,3 +1177,12 @@ oneOrMoreHelp spaces p vs =
         , P.succeed ()
             |> P.map (always (P.Done (List.reverse vs)))
         ]
+
+
+{-| Taken from dmy/elm-pratt-parser, made to accept the operator parser result
+-}
```
Janiczek, line 1183: Can you expand on this description? Eg. in the `foo as bar` example, I'd imagine `as` being the operator, but it seems like you're saying `bar` is the operator parser result?

tests/ParserTest.elm
```diff
@@ -964,6 +964,72 @@ expr =
...
+              , [ ( "simple case"
+                  , "case True of _ -> True"
+                  , Just
+                        (Case (Bool True)
+                            [ { pattern = PAnything, body = Bool True }
+                            ]
+                        )
+                  )
+                , ( "multiline case"
+                  , "case 21 of\n31 -> True\n5 -> True\n_ -> False"
```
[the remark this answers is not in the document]
harrysarson: I have never understood elms indentation rules, except that if you press tab enough times and then run elm-format it usually works :laughing:

src/Stage/Desugar.elm
```diff
@@ -196,6 +196,118 @@ desugarExpr modules thisModule located =
...
+            Result.map
+                (\pattern ->
+                    Located.replaceWith
+                        (fn pattern)
+                        located
+                )
+
+        map2 fn =
+            Result.map2
+                (\pttrn1 pttrn2 ->
```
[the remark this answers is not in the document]
harrysarson: C sees fit to shorten `memcopy` to `memcpy` which saves one less character than shortening `pattern1` to `pttrn1`.

Not that C sets a good example on API in [any way](https://en.cppreference.com/w/c/io/gets):

> The gets() function does not perform bounds checking, [..]  It cannot be used safely [..]. For this reason, the function has been deprecated in [...] C99 [..] and removed altogether in the C11 standard.

pablohirafuji:
Hey @Janiczek, thank you for the review!

> Sorry for taking so long to review it. I had busy weekends lately

No worries 😄 

> the missing emitter (do we want to do it here or let it be TODO for now?)

I will try to figure out in this PR.

> comments shouldn't be ignored by the parser but kept in the AST

I will remove the comments parsing/ignoring from this PR. I can do it in another PR, after finishing this one if no one is working on it.


Ok, I will work on the requested changes, thanks!

src/Stage/InferTypes/GenerateEquations.elm
pablohirafuji: You are right, we can say "It's a record and have at least these keys". I will take a look and see if I have the knowledge to express it or, at least, improve the comment.

tests/ParserTest.elm
```diff
@@ -964,6 +964,72 @@ expr =
...
+                    , "((arg1, arg2), 435.4) ->"
+                    , " False"
+                    , "[_, 45, (67.7)] ->"
+                    , " False"
+                    , "fst :: snd :: tail ->"
+                    , " False"
+                    , "({ count } as alias1) as alias2 ->"
+                    , " False"
+                    ]
+                        |> String.join "\n"
```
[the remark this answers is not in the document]
pablohirafuji: That's a good idea, even easier to read. I will implement it!

src/Stage/InferTypes/GenerateEquations.elm
Janiczek: I'm also guessing the whole type will be the same as that delegated type (the var).
`equals type_ varType`

pablohirafuji:
Working on the emit phase.

Janiczek:
@pablohirafuji 

Looking great!
So, feel free to make a line of where to stop. For the `case ... of` to be complete, I feel like this is missing - but we can do it in other PRs:

- desugaring: thinking / reading up on whether `case...of` pattern matching can be desugared to some form of `if...else`. I believe it can't because we'd need some unwrapping of custom types. (ie. not all `case...of` can be converted to `if...else` in Elm)

- type inference tests
- type inference generating equation TODOs

- optimizer tests
- some optimizations (`case bool of True / False` -> `if...else` seems like an easy one? more generally literals? Unsure, I didn't think about this too much in depth)

- emit tests
- emitting

If you want any rubber-duck help with any of these, just ask :)
And again, we don't need to do all these here.

harrysarson:
> desugaring: thinking / reading up on whether case...of pattern matching can be desugared to some form of if...else. I believe it can't because we'd need some unwrapping of custom types. (ie. not all case...of can be converted to if...else in Elm)

It might be better to do the desuger the other way I.e. convert:

```elm
if bool then
   x
else
  y
```

Into


```
case condition of
  True -> x
  False -> y
```

pablohirafuji:
The emit phase is a big task. Has a lot of requirements that I currently doesn't have the knowledge. I would like to focus on the parsing stage for now, while it is still fresh on my mind, if you don't mind.

Can we close this PR and work on the parsing of the comments next?

Janiczek:
> Can we close this PR and work on the parsing of the comments next?

@pablohirafuji Absolutely! I welcome all the help with parsing.

Would you like to add yourself to the contributors table at the bottom of the README?

After we figure that out, I'm gonna merge this PR :)

pablohirafuji:
> Would you like to add yourself to the contributors table at the bottom of the README?

Yes, thanks! I also checked the `parser tests` and `parse` columns of `case...of` roadmap row (and aligned some misaligned rows).

PR #105, opened 2020-06-30
add failing test with two source directories

harrysarson:
I was hoping to submit a patch to allow the compiler
to work correctly with multiple source directories.

Unfortunately, it turned out to be too hard so I am going
to pick a different fight. I added some integration tests
before I gave up though.

I am not sure what our policy is about adding failing tests, 
feel free to reject the failing one!

- add module import test
- add failing test with two source dirs


Janiczek:
I'll look at this PR over today / tomorrow, thanks!

Failing tests are fine for now :) Perhaps that will change as we stabilize, but for now they're OK.

integration-tests/cli/two-source/test.test.js
```diff
@@ -0,0 +1,10 @@
+const test = require('ava');
+const {exec} = require('../../setup');
+
+test.failing(
+    'parser: able to import variable from module',
```
Janiczek, line 5: Perhaps let's not mark it as `test.failing` ("meant to fail") but `test` (meant to work, and failures should be reported)?

Also I'd write something more in the message about there being two source directories, given that's in the test folder name and that's what the Elm source files are actually doing.

integration-tests/parser/import/test.test.js
```diff
@@ -0,0 +1,10 @@
+const test = require('ava');
+const {exec} = require('../../setup');
+
+test(
+    'cli: able to have two source directories',
```
Janiczek, line 5: This test case seems to say what the other one should have said; if that's the case and they're switched accidentally, what should this one be about?

integration-tests/cli/two-source/test.test.js
harrysarson:
```suggestion
    'cli: supports multiple source directories',
```

integration-tests/parser/import/test.test.js
harrysarson:
```suggestion
    'parser: able to import variable from module',
```

oops!

harrysarson:
> Perhaps let's not mark it as test.failing ("meant to fail") but test (meant to work, and failures should be reported)?

But then we get a nasty red cross against the repository

integration-tests/cli/two-source/test.test.js
```diff
@@ -0,0 +1,10 @@
+const test = require('ava');
+const {exec} = require('../../setup');
+
+test.failing(
```
Janiczek, line 4: So, again, should this be `test.failing`? (Sorry if I missed an answer somewhere.)
IMHO we can use `test` for things that should work and `test.failing` for things that should fail (compiler errors etc.).
This seems like something that should work (but perhaps doesn't yet, I don't know and I'm AFK :D ) so I'd expect `test`

integration-tests/parser/import/test.test.js
```diff
@@ -0,0 +1,10 @@
+const test = require('ava');
+const {exec} = require('../../setup');
+
+test(
+    'parser: able to import variable from module',
```
Janiczek, line 5: IMHO this is more about desugaring (finding correct module reference for each used value/type) than parser. WDYT?

integration-tests/cli/two-source/test.test.js
harrysarson: See https://github.com/elm-in-elm/compiler/pull/105#issuecomment-652056367.

We don't want failing tests on master do we? I am definitely happy to drop this test from the PR. I added it when I thought I could also add a fix for it.

Janiczek: Ah, thanks, apparently I'm blind :)
Well I'd rather mark it as "should not fail but fails".

> Failing tests are fine for now :) Perhaps that will change as we stabilize, but for now they're OK.

So I guess either 
1. mark it as `test` and add a fix so that it passes
2. mark it as `test` and let it be :x: in master (we'll fix it later)
3. remove the test (perhaps add a TODO somewhere?)

But I don't like pretending it _should_ fail.

integration-tests/parser/import/test.test.js
harrysarson: I agree

integration-tests/cli/two-source/test.test.js
harrysarson: My understanding (see [ava docs](https://github.com/avajs/ava/blob/f72fab40160a413f4d7ab0208979fe3e721198ea/docs/01-writing-tests.md#failing-tests)) is that the `.failing` does _not_ mark a test as "meant to fail" but instead "currently failing and needs to be fixed". In this way `.failing` works the same as `.skip` but ava checks that the test does actually fail and so you get prompted to remove the `.failing` if you fix the issue.

harrysarson: Option 1. is definitely preferable to me than option 2. otherwise

Janiczek: Hm, then we should change our usages of `test.failing` for something else, because they all should fail. I'm gonna read the docs...

Janiczek: I guess the (expected) compiler failures should be somehow test successes. We'll need to rethink this a bit.

Anyways, with the newly acquired knowledge, `test.failing` is fine for the "not yet implemented" test. :+1:

harrysarson: The only other `.failing` test is currently https://github.com/elm-in-elm/compiler/search?q=failing which is also a  "not yet implemented" test so I think we may be okay :)

Janiczek: My bad, I'm in the type-annotations branch in my head... I was using the integration tests wrong there. Added about 7 failing tests that are implemented fully and supposed to fail. :see_no_evil:

harrysarson: I put some (hopefully helpful) comments over in that PR :)

PR #107, opened 2020-07-01
add if-then-else integration test

harrysarson:
This pull request adds an integration test that should pass but does not. However, when I revert the changes in #65 the integration test does pass! (I am not in anyway suggesting we should revert #65, the purpose of the revert is to demonstrate that this PR regressed this test.) Somewhere the typechecking goes wrong but I am not sure how to debug this further.

The compiler output (when I haven't reverted #65) is:

```
Compiled in DEV mode. Follow the advice at https://elm-lang.org/0.19.1/optimize for better performance and smaller assets.
---------------------------
-- STARTING THE COMPILER --
---------------------------
Main.main: Value { expression = Plus (Call { argument = Bool True, fn = Var { name = "f", qualifiedness = PossiblyQualified Nothing } }) (Call { argument = Bool False, fn = Var { name = "f", qualifiedness = PossiblyQualified Nothing } }), typeAnnotation = Nothing }
Main.f: Value { expression = Lambda { arguments = ["x"], body = If { else_ = Int 2, test = Argument "x", then_ = Int 18 } }, typeAnnotation = Nothing }


---------------------------
-- COMPILER ERROR ---------
---------------------------
The types `Bool` and `Int` don't match.
```


Janiczek:
Thanks @harrysarson!

As a starting point, it's possible there is some faulty logic regarding creating new `SubstitutionMap`s and combining them. I know I had some TODO about it but can't recall (due to the the PR being done over months instead of days) if it has been solved well or not.

I'll take a look. Also to be transparent, I start debugging this inside `Stage.InferTypes` by `Debug.log`-ing the things inside `inferExpr` there.

Janiczek:
Putting your failing test source code into `example-project/Main.elm` and running `make`, I get:
```
---------------------------
-- STARTING THE COMPILER --
---------------------------
Main.main: Value { expression = Plus (Call { argument = Bool True, fn = Var { name = "f", qualifiedness = PossiblyQualified Nothing } }) (Call { argument = Bool False, fn = Var { name = "f", qualifiedness = PossiblyQualified Nothing } }), typeAnnotation = Nothing }
Main.f: Value { expression = Lambda { arguments = ["x"], body = If { else_ = Int 2, test = Argument "x", then_ = Int 18 } }, typeAnnotation = Nothing }
---------------------: ()
expr with ids: (Lambda { argument = "x", body = (If { else_ = (Int 2,Id 2), test = (Argument "x",Id 0), then_ = (Int 18,Id 1) },Id 3) },Id 4)
new substitution map: Ok (SubstitutionMap (Dict.fromList [(0,Type Bool),(1,Id 2),(2,Id 3),(3,Type Int),(4,Type (Function { from = Id 5, to = Id 3 })),(5,Id 0)]))
---------------------: ()
expr with ids: (Plus (Call { argument = (Bool True,Id 1), fn = (Var { module_ = "Main", name = "f" },Id 0) },Id 2) (Call { argument = (Bool False,Id 4), fn = (Var { module_ = "Main", name = "f" },Id 3) },Id 5),Id 6)
new substitution map: Err (TypeMismatch (Type Bool) (Type Int),SubstitutionMap (Dict.fromList [(0,Type Bool),(1,Id 2),(2,Id 3),(3,Type Int),(4,Type (Function { from = Id 5, to = Id 3 })),(5,Id 0)]))

---------------------------
-- COMPILER ERROR ---------
---------------------------
The types `Bool` and `Int` don't match.
```
which is a bit hard to parse (we might develop some Graphviz exports for these debug dumps?) but I've parsed it into this:
IMG_20200702_010627

The `f` seems fine (`Bool -> Int`) and then we see IDs generated for `main` but they also start from 0, which can't be good if we're reusing the `SubstitutionMap` from `f`. I'll be following this hunch :upside_down_face:

Janiczek:
Also: feel free to throw away the revert commit from this PR - I've seen the proof that #65 has broken this and am sending you an ACK :)

Janiczek:
If you'll remove the revert and merge `origin/master` to your branch (basically, https://github.com/elm-in-elm/compiler/commit/f7384de489342fa6d7efb8f39839436277a5aa77), you should find the test passes now.

IMG_20200702_014935

The hunch was correct - the `f`'s and `main`'s IDs are now distinct and so the SubstitutionMap doesn't have a problem with `main`'s IDs.

harrysarson:
Makes sense!  Nice fix too. I can hopefully remove the revert tonight 👍

harrysarson:
rebased!

PR #108, opened 2020-07-02
parsin

harrysarson:
Most of these commits a) add a integration-test b) tweak
the parser so that integration-test passes.

I also fix a bug where the snapshot from a test is not 
wrapped in a object in the correct way and so is silently
skipped. There is now an assertion to check the snapshot
is correctly wrapped.


integration-tests/parser/indentation/pipe-needing-indentation/test.test.js.md
```diff
@@ -0,0 +1,27 @@
...
+    'elm-in-elm -m src/Main.elm'
+
+## Stderr
+
+    `Compiled in DEV mode. Follow the advice at https://elm-lang.org/0.19.1/optimize for better performance and smaller assets.␊
+    ␊
+    ---------------------------␊
+    -- COMPILER ERROR ---------␊
+    ---------------------------␊
+    Couldn't find the value `main` in the main module given to the compiler!␊
```
Janiczek, line 18: Not ideal, huh? :smile:

harrysarson: True! I didn't actual take care to read this message I was too concerned about the exit code.

src/Stage/Parse/Parser.elm
```diff
@@ -1236,8 +1240,7 @@ patternTuple config =
 
 spacesOnly : Parser_ ()
 spacesOnly =
-    P.succeed identity
-        |= P.chompWhile ((==) ' ')
+    P.chompWhile ((==) ' ')
```
Janiczek, line 1243: :man_facepalming: Thanks :)

integration-tests/parser/indentation/pipe-needing-indentation/test.test.js.md
harrysarson: With `P.End` in `module_` I get:

```
---------------------------
-- STARTING THE COMPILER --
---------------------------

---------------------------
-- COMPILER ERROR ---------
---------------------------
Parse problem: ExpectingEnd
  --> src/Main.elm:6:1
   | 
6  | | Nothing
   | ^ ExpectingEnd
```

harrysarson: The issue is we have

```elm
                |= P.oneOf
                    [ P.succeed (\newCtor -> P.Loop (newCtor :: reversedCtors))
                        |. onlyIndented (P.token (P.Token "|" ExpectingPipe))
                        |. P.spaces
                        |= onlyIndented constructor
                    , P.succeed (P.Done (List.reverse reversedCtors))
                    ]
    ```

and we do not commit to the "|".

Comment (harrysarson) [reply integration-tests/parser/indentation/pipe-needing-indentation/test.test.js.md:18]: and so we need to rework the `onlyIdented` function to commit if the parser would succeed (regardless of indentation).

```elm
{-| Fail if current column <= 1 (these are 1-based, so 1 is leftmost.)
-}
onlyIndented : Parser_ a -> Parser_ a
onlyIndented parser =
    P.succeed identity
        |. checkIndent (\_ column -> column > 1) ExpectingIndentation
        |= parser
```

@Janiczek do you know how we would do that. Relevant: https://github.com/elm/parser/blob/master/semantics.md

harrysarson: We talked about this here: https://discord.com/channels/578305644780716039/578306156972212224/728362159288877117

> need to rework the onlyIdented function to commit if the parser would succeed (regardless of indentation).

This is false! The problem **isn't** that the custom type parser did not commit to the "|". Instead the problem was that the `declarations` parser was happy to stop parsing and return success as soon as it encountered an error!

Janiczek:
:heart:

PR #111, opened 2020-07-08
refuse to parse operators unless they are indented

harrysarson:


src/Stage/Parse/Parser.elm
```diff
@@ -548,9 +548,21 @@ expr =
...
+            , PP.infixLeft 1
+                (checkIndent (<) ExpectingIndentation
+                    |> P.andThen (\() -> P.symbol (P.Token "+" ExpectingPlusOperator))
+                )
+                (Located.merge Plus)
+            , PP.infixRight 1
+                (checkIndent (<) ExpectingIndentation
+                    |> P.andThen (\() -> P.symbol (P.Token "::" ExpectingConsOperator))
+                )
+                (Located.merge Cons)
```
harrysarson, line 565: @Janiczek this new code is _more correct_ but less readable and less DRY. Do you have any thoughts on how it can be tidied?

harrysarson:
todo: remove https://github.com/elm-in-elm/compiler/pull/111/files#diff-581df35285baeb14ec82a2d6d5659052R540

src/Stage/Parse/Parser.elm
Janiczek: Soon we'll remove these three cases and replace them with a generic operator. So there will be only one case. Does that sound reasonable?

integration-tests/parser/indentation/expr-needing-indentation/src/Main.elm
```diff
@@ -0,0 +1,6 @@
+module Main exposing (MyMaybe)
+
+
+main =
+  1
++ 7
```
Janiczek, line 6: :+1: I think we can make this test better!

It tests the "trivial" case where indentation is 0; I'd like to test it (perhaps in a separate test?) where the indentation is different from 0. Eg. something like:`let..in` which sets the indentation to the binding level (where `x` is).

FAIL :x:
```elm
main =
    let
        x =
            1
        + 2
    in
    x
```

PASS :heavy_check_mark: 
```elm
main =
    let
        x =
            1
         + 2
    in
    x
```
(Tested on official compiler in Ellie)

harrysarson: The problem is that for failing tests we can only have one failure per file so testing multiple failure modes quickly gets a bit much

harrysarson: We should for example test other operators such as `++`

src/Stage/Parse/Parser.elm
harrysarson: Sounds reasonable! I guess this applies to my colment below too!

harrysarson:
CI fails but it fails in master too!

PR #115, opened 2020-08-15
Added GLSL parser

decioferreira:
An initial implementation for the WebGL Shader (GLSL) parser.

### DIFF
--- a/src/Elm/AST/Typed.elm
+++ b/src/Elm/AST/Typed.elm
@@ -84,6 +84,7 @@ type Expr_
     | Tuple3 LocatedExpr LocatedExpr LocatedExpr
     | Record (Dict VarName (Binding LocatedExpr))
     | Case LocatedExpr (List { pattern : LocatedPattern, body : LocatedExpr })
+    | Shader String
 
 
 type alias LocatedPattern =
@@ -214,6 +215,9 @@ recurse fn locatedExpr =
                                     }
                                 )
                                 branches
+
+                    Shader shader ->
+                        Shader shader
             )
 
 
@@ -307,6 +311,9 @@ recursiveChildren fn locatedExpr =
 
         Case e branches ->
             fn e ++ List.fastConcatMap (.body >> fn) branches
+
+        Shader _ ->
+            []
 
 
 mapExpr : (Expr_ -> Expr_) -> LocatedExpr -> LocatedExpr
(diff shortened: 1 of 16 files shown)


src/Elm/AST/Typed.elm
```diff
@@ -214,6 +215,9 @@ recurse fn locatedExpr =
                                     }
                                 )
                                 branches
+
+                    Shader shader ->
+                        Shader shader
```
Janiczek, line 220: We could do `expr` instead of explicitly writing the value out, but that's just a small nitpick :)

src/Stage/Parse/Parser.elm
```diff
@@ -1450,6 +1451,20 @@ case_ config =
...
+shader config =
+    let
+        startToken =
+            "[glsl|"
+    in
+    P.succeed ()
+        |. P.symbol (P.Token startToken ShaderProblem)
+        |. P.chompUntil (P.Token "|]" ShaderProblem)
+        |> P.getChompedString
+        |> P.map (String.dropLeft (String.length startToken) >> Frontend.Shader)
```
Janiczek, line 1464: I believe we could get rid of the `dropLeft` postprocessing by instead chomping only after that prefix.

I'm skipping boilerplate because I'm writing this on my phone, but the idea would be:

```elm
P.succeed Frontend.Shader
|. P.symbol "[glsl|"
|= P.getChompedString (P.chompUntil "|]")
|. P.symbol "|]"
|> located
```

tests/InferTypesFuzz.elm
```diff
@@ -406,6 +406,10 @@ shrinkExpr expr =
             -- TODO take a stab at this? Do we actually even generate these?
             nope
 
+        CanonicalU.Shader _ ->
+            -- TODO take a stab at this? Do we actually even generate these?
```
Janiczek, line 410: I wonder if we would have to make well-defined GLSL programs here. Let's leave it as it is right now :+1:

Janiczek:
Looks good so far!
Please let's add one more example to ParserTest with a correct GLSL program (can be very simple but probably should use the varyings etc. that change the type of the shader).

I'm curious to see how the shader type turns out! Don't hesitate to ask on Discord if you want help with generating the type equations :)

(Type inference tests will be the most interesting and important part here I believe.)

decioferreira:
added two more examples to the shader parser tests, based of https://github.com/elm-explorations/webgl/blob/81527c23ba21d9bf92f34151a6a4c40a68460e1f/examples/Triangle.elm#L77-L105

PR #116, opened 2020-10-06
add type annotations

harrysarson:
As suggested by new elm vscode release

### DIFF
--- a/tests/Shrink/Extra.elm
+++ b/tests/Shrink/Extra.elm
@@ -13,6 +13,7 @@ The `LazyList a` type used by shrinkers is not exposed outside `elm-explorations
     map2 : (a -> b -> c) -> LazyList a -> LazyList b -> LazyList c
 
 -}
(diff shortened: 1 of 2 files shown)


tests/Shrink/Extra.elm
```diff
@@ -28,6 +29,7 @@ The `LazyList a` type used by shrinkers is not exposed outside `elm-explorations
     map3 : (a -> b -> c -> d) -> LazyList a -> LazyList b -> LazyList c -> LazyList d
 
 -}
+map3 : (a -> b -> c -> d) -> LazyList a -> LazyList b -> LazyList c -> LazyList d
```
Janiczek, line 32: Hi @harrysarson! Where does this `LazyList` get imported from?

harrysarson: ah, very good point!

Maybe is a bug in the language server type inference

harrysarson: Is just annoying because now I get squigly lines in my IDE and I cannot work out how to turn them off!

